PR #4, opened 2018-10-20
allow FreeBSD 13

lrosenman:
and how do we make new tarballs?

### DIFF
--- a/Configure
+++ b/Configure
@@ -1626,7 +1626,7 @@ kernel generation process.
 	;;
       13*)
 	LSOF_TSTBIGF=" "
-	LSOF_VERS=12000
+	LSOF_VERS=13000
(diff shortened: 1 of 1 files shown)


masatake:
As I wrote ago, I think we should start new things after cleaning up the source tree.
I made this repository for the cleaning-up stage. If you don't remember the plan I designed, let me know.
I thought we agreed on the plan. I took much my time for designing the plan. So, I'm disappointed with this pull request.
However, extending the life of lsof is much important than my sentiment. so, please, do as you want.
One having the time works on lsof may be better than do nothing.

masatake:
I would like you to read the original plan.

masatake:
I will merge this in 4.92.2.

Issue #38, opened 2019-05-09
lsof +fg doesn't recognize O_CLOEXEC in file flags on Linux


barsnick:
lsof +fg doesn't recognize O_CLOEXEC in file flags on Linux. It displays this flag as "0x80000". It should probably be displayed as "CX".

This was pointed out in [this RedHat bugzilla issue](https://bugzilla.redhat.com/show_bug.cgi?id=1599606).

I assume Linux's O_CLOEXEC is identical to other platforms' POF_CLOEXEC. I took a stab at this. Will post a PR.

barsnick:
This shows the change in behavior with this patch versus vanilla lsof:
```
$ diff -ub <(lsof +fg -p $$) <(./lsof +fg -p $$)
--- /dev/fd/63  2019-05-09 15:03:23.820280586 +0200
+++ /dev/fd/62  2019-05-09 15:03:23.822280571 +0200
@@ -13,5 +13,5 @@
 bash    631 barsnick    0u   CHR         RW,AP  136,2       0t0       5 /dev/pts/2
 bash    631 barsnick    1u   CHR         RW,AP  136,2       0t0       5 /dev/pts/2
 bash    631 barsnick    2u   CHR         RW,AP  136,2       0t0       5 /dev/pts/2
-bash    631 barsnick    3r   REG    LG,0x80000  253,0   8406312 1049099 /var/lib/sss/mc/passwd
-bash    631 barsnick  255u   CHR RW,AP,0x80000  136,2       0t0       5 /dev/pts/2
+bash    631 barsnick    3r   REG     CX,LG  253,0   8406312 1049099 /var/lib/sss/mc/passwd
+bash    631 barsnick  255u   CHR  RW,AP,CX  136,2       0t0       5 /dev/pts/2
```

I assume (and hope) this fix is valid and correct.

masatake:
See https://github.com/lsof-org/lsof-linux/search?q=O_CLOEXEC&type=Commits

masatake:
I forget to write "Thank you."

masatake:
I have some patches about open flags at lsof-org/lsof-linux repo.
```
  /home/yamato/var/lsof-linux/patches/oflags:
  total used in directory 56 available 701427864
  drwxrwxr-x. 2 yamato yamato 4096 May  8 21:10 .
  drwxr-xr-x. 8 yamato yamato 8192 May  9 19:10 ..
[...]

PR #39, opened 2019-05-09
Add support for O_CLOEXEC flag on Linux

barsnick:
Closes lsof-org/lsof#38


masatake:
Do you think `#if	defined(O_CLOEXEC)` is needed here?

barsnick:
I would say so. I don't have any old Linux kernel anymore, but the docs say
```since Linux 2.6.23```. There used to be some yet-available legacy LTS distributions which still used 2.6.16, but those days may be over. Consider non-availability of this define an "old dialect" - it might just happen.

masatake:
I looked into strace to know how strace handles these case.
strace also use "#if defined O_CLOEXEC".
I will add the condition to  my patch.
Thank you.

masatake:
#40 does the same.

PR #44, opened 2019-05-13
Implement endpoint feature for inet4 socket used  for IPC

masatake:


masatake:
Ported from lsof-linux repo.

PR #47, opened 2019-05-13
Implementat [-+]E option for locally used IPv6 sockets 

masatake:


masatake:
The coverage is reduced because travis-ci disables ipv6 stack.

PR #51, opened 2019-05-22
support FreeBSD-current

lrosenman:
FreeBSD -CURRENT (HEAD) needs to include <sys/_lock.h>.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,17 +1,18 @@
-# freebsd13_task:
+#freebsd13_task:
 #   freebsd_instance:
-#     image: freebsd-13-0-current-amd64-v20190503
+#     image: freebsd-13-0-current-amd64-v20190517
 #   timeout_in: 120m
-#   install_script: pkg install -y gcc git bash
+#   install_script: pkg install -y git bash
 #   script:
-#     - ./Configure freebsd
+#     - ./Configure -n freebsd
 #     - make
+#     - bash ./check.bash freebsd
 
 freebsd12_task:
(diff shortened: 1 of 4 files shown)


masatake:
Thank you for updating 00DIST.
You may also want to add an entry at the end of 00DIST.

masatake:
A very minor issue:
The commit log "add comment in 00DISR" should be "add comment in 00DIST".

lrosenman:
> A very minor issue:
> The commit log "add comment in 00DISR" should be "add comment in 00DIST".

I'm not git-knowledgeable to fix it :(.

masatake:
About "move to newer FreeBSD 13 image, remove installing gcc, as its not needed ", could you split the
commit into two smaller commits: one is about FreeBSD13 new image, and another is about gcc.
I think it will take time to fix "FreeBSD13 new image" on cirrus issue. If the commits are split, we can merge the commit about gcc only in soon.

lrosenman:
@masatake thank you for the ideas.  I've fixed them, I think.  Please do let me know if we need more to merge and make a 4.93.3 release.  (I've got the Configure and dialects/freebsd/dlsof.h changed patched into the FreeBSD port, so there's no rush).

masatake:
I would like to rearrange your commit a bit.
I will make a pull request. So, please review it.

I would like to introduce a new convention about writing commit log.
I would like to introduce [freebsd] and  [linux] as a prefix.

masatake:
See #52.

lrosenman:
closing in deference to #52

PR #63, opened 2019-06-23
Allow to specify max iteration count to -r option

masatake:
e.g.

	$ ./lsof -r1c3 -p 1

This command line lists files currently opend by the process having 1
as pid 3 times with 1 sec interval.

This enhancement helps us to run lsof under valgrind.

### DIFF
--- a/00DIST
+++ b/00DIST
@@ -4999,5 +4999,8 @@ July 14, 2018
                 [FreeBSD] update to include <sys/_lock.h> on recent -CURRENT
 		since it is no longer implicitly included via header pollution.
 
+		Enhanced -r option. With `c<N>' specifier, lsof can stop itself
+		when the number of iterations reaches at <N>.
+
 Masatake YAMATO <yamato@redhat.com>, a member of the lsof-org team at GitHub
 May 8, 2019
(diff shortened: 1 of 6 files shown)


masatake:
At least, a test case is needed.

masatake:
A test case is added.

masatake:
Could you review the change for Lsof.8 ?
```
diff --git a/Lsof.8 b/Lsof.8
index 4541219..4a05d6e 100644
--- a/Lsof.8
+++ b/Lsof.8
@@ -1373,7 +1373,7 @@ Inhibiting the conversion may make
 run a little faster.
 It is also useful when port name lookup is not working properly.
 .TP \w'names'u+4
-.BI +|\-r " [t[m<fmt>]]"
+.BI +|\-r " [t[c<N>][m<fmt>]]"
 puts
 .I lsof
 in repeat mode.
@@ -1388,6 +1388,10 @@ the prefix to the option.
 If the prefix is a `\-', repeat mode is endless.
 .I Lsof
 must be terminated with an interrupt or quit signal.
+`c<N>' is for specifying the limits of repeating; if the number of
+iterations reaches at `<N>',
+.I Lsof
+stops itself.
 .IP
 If the prefix is `+', repeat mode will end the first cycle no open files
 are listed \- and of course when
```

masatake:
@lrosenman, thank you.

PR #67, opened 2019-07-08
Build man page for linux using soelim. 

greenius:
Also renamed source man page to avoid potential problems building on case insensitive file systems which would treat Lsof.8 and lsof.8 as the same file (causing problem with adding lsof.8 to .gitignore)


masatake:
Thank you.  I wonder how large the impact of this change is on the other dialect like freebsd.

masatake:
Ideally, converting Lsof.8 to lsof.8 should be done in "make dist".
Linux allows users to run lsof without root-privilege. So the man page should be lsof.1.

I have to revise the process for releasing.

lrosenman:
currently. the FreeBSD port makefile does:
` (cd ${WRKSRC} && /usr/bin/soelim Lsof.8 > ${WRKSRC}/lsof.8)`

jiegec:
Already done in current version.

Issue #69, opened 2019-07-19
Fails to build on FreeBSD PowerPC64


ghost:
Hi, sysutils/lsof fails to build on FreeBSD / PowerPC64 with the following error:

cc -o lsof  -pipe -fstack-protector-strong -fno-strict-aliasing -DNEEDS_BOOL_TYPEDEF -DHASTASKS -DHAS_PAUSE_SBT -DHAS_DUP2 -DHAS_CLOSEFROM -DHASEFFNLINK=i_effnlink -DHASF_VNODE -DHAS_FILEDESCENT -DHAS_TMPFS -DHASWCTYPE_H -DHASSBSTATE -DHAS_KVM_VNODE -DHAS_UFS1_2 -DHAS_NO_IDEV -DHAS_VM_MEMATTR_T -DNEEDS_DEVICE_T -DHAS_CDEV2PRIV -DHAS_NO_SI_UDEV -DHAS_SYS_SX_H -DHASFUSEFS -DHAS_ZFS -DHAS_V_LOCKF -DHAS_LOCKF_ENTRY -DHAS_NO_6PORT -DHAS_NO_6PPCB -DNEEDS_BOOLEAN_T -DHAS_SB_CCC -DHAS_FDESCENTTBL -DFREEBSDV=13000 -DHASFDESCFS=2 -DHASPSEUDOFS -DHASNULLFS -DHASIPv6 -DHASUTMPX -DHAS_STRFTIME -DLSOF_VSTR=\"13.0-CURRENT\" -I/usr/src/sys -O2 dmnt.o dnode.o dnode1.o dnode2.o dproc.o dsock.o dstore.o  arg.o main.o misc.o node.o print.o proc.o store.o usage.o  util.o -L./lib -llsof  -lkvm
dnode2.o:(.text+0x190): undefined reference to `abd_zero_off'
dnode2.o:(.text+0x1c8): undefined reference to `abd_cmp_buf_off'
dnode2.o:(.text+0x200): undefined reference to `abd_copy_to_buf_off'
dnode2.o:(.text+0x238): undefined reference to `abd_copy_from_buf_off'
dnode2.o:(.text+0x274): undefined reference to `abd_copy_off'
*** Error code 1

Stop.
make[1]: stopped in /wrkdirs/usr/ports/sysutils/lsof/work/lsof-4.93.2
*** Error code 1

Stop.
make: stopped in /usr/ports/sysutils/lsof

[...]

PR #70, opened 2019-07-25
[FreeBSD] add <stdbool.h> to dlsof.h for refcount(9) issue

lrosenman:


lrosenman:
@masatake assuming this meets your approval, can we create a new release?  Thanks.

lrosenman:
if we want to make the minimal change, we can just include <stdbool.h> instead of <sys/param.h>

masatake:
We should add the name of dialect as a prefix for the commit log. How about following?

```
[freebsd] add <sys/param.h> to dlsof.h for refcount(9)

Close #70 
```

You should update 00DIST.

Could you see #63?

masatake:
I also think we should make a release.
However, I would like to revise the release procedure before releasing.
See https://github.com/lsof-org/lsof/pull/67.

Currently, I use the automatic(?) archive feature that GitHub provides.
To solve #67, we have to have a script making an archive, and have to run it before making a release.
It means we have to inspect the scripts Vic wrote...It will take time.

I will try to find the time to merge #67 in August. Even if I cannot find the time, let's make a release around early September.

masatake:
Let's squash the two change into one.

masatake:
If you really think, we should make a release as soon as possible. I will not object to it.

lrosenman:
I'd like to make a release and feel free to squash it on merge.

masatake:
@thexcoder, could you try @lrosenman's branch?

masatake:
@lrosenman, how can we know #70 is fixed with this change?

lrosenman:
I've compiled it on my -CURRENT system.  Or you can wait for the next -CURRENT snapshots this weekend in cirrus.

masatake:
??? I assumed the issue can be reproduced only on PowerPC.
Did you buy an oldmac?

masatake:
(I will go out for a day.)

lrosenman:
Oh, there is a NEW issue (even on i386) with refcount(9) api calls getting the following:
`
===========================================================================
=======================<phase: build          >============================
===>  Building for lsof-4.93.2_2,8
(cd lib; /usr/bin/make DEBUG="-O2" CFGF="-pipe -fstack-protector-strong -fno-strict-aliasing -DNEEDS_BOOL_TYPEDEF -DHASTASKS -DHAS_PAUSE_SBT -DHAS_DUP2 -DHAS_CLOSEFROM -DHASEFFNLINK=i_effnlink -DHASF_VNODE -DHAS_FILEDESCENT -DHAS_TMPFS -DHASWCTYPE_H -DHASSBSTATE -DHAS_KVM_VNODE -DHAS_UFS1_2 -DHAS_NO_IDEV -DHAS_VM_MEMATTR_T -DNEEDS_DEVICE_T -DHAS_CDEV2PRIV -DHAS_NO_SI_UDEV -DHAS_SYS_SX_H -DHASFUSEFS -DHAS_ZFS -DHAS_V_LOCKF -DHAS_LOCKF_ENTRY -DHAS_NO_6PORT -DHAS_NO_6PPCB -DNEEDS_BOOLEAN_T -DHAS_SB_CCC -DHAS_FDESCENTTBL -DFREEBSDV=13000 -DHASFDESCFS=2 -DHASPSEUDOFS -DHASNULLFS -DHASIPv6 -DHASUTMPX -DHAS_STRFTIME -DLSOF_VSTR=\"13.0-CURRENT\"")
cc   -pipe -fstack-protector-strong -fno-strict-aliasing -DNEEDS_BOOL_TYPEDEF -DHASTASKS -DHAS_PAUSE_SBT -DHAS_DUP2 -DHAS_CLOSEFROM -DHASEFFNLINK=i_effnlink -DHASF_VNODE -DHAS_FILEDESCENT -DHAS_TMPFS -DHASWCTYPE_H -DHASSBSTATE -DHAS_KVM_VNODE -DHAS_UFS1_2 -DHAS_NO_IDEV -DHAS_VM_MEMATTR_T -DNEEDS_DEVICE_T -DHAS_CDEV2PRIV -DHAS_NO_SI_UDEV -DHAS_SYS_SX_H -DHASFUSEFS -DHAS_ZFS -DHAS_V_LOCKF -DHAS_LOCKF_ENTRY -DHAS_NO_6PORT -DHAS_NO_6PPCB -DNEEDS_BOOLEAN_T -DHAS_SB_CCC -DHAS_FDESCENTTBL -DFREEBSDV=13000 -DHASFDESCFS=2 -DHASPSEUDOFS -DHASNULLFS -DHASIPv6 -DHASUTMPX -DHAS_STRFTIME -DLSOF_VSTR="13.0-CURRENT" -I/usr/src/sys -O2 -c ckkv.c -o ckkv.o
In file included from ckkv.c:43:
In file included from ./../lsof.h:221:
In file included from ./../dlsof.h:406:
In file included from /usr/src/sys/sys/file.h:44:
/usr/src/sys/sys/refcount.h:65:12: error: use of undeclared identifier 'false'
                        return (false);
                                ^
/usr/src/sys/sys/refcount.h:68:12: error: use of undeclared identifier 'true'
                        return (true);
                                ^
/usr/src/sys/sys/refcount.h:81:11: error: use of undeclared identifier 'false'
                return (false);
                        ^
/usr/src/sys/sys/refcount.h:90:10: error: use of undeclared identifier 'true'
        return (true);
                ^
/usr/src/sys/sys/refcount.h:106:12: error: use of undeclared identifier 'false'
                        return (false);
                                ^
/usr/src/sys/sys/refcount.h:108:12: error: use of undeclared identifier 'true'
                        return (true);
                                ^
/usr/src/sys/sys/refcount.h:121:12: error: use of undeclared identifier 'false'
                        return (false);
                                ^
/usr/src/sys/sys/refcount.h:123:12: error: use of undeclared identifier 'true'
                        return (true);
                                ^
8 errors generated.
`

and this fixes THAT issue.  The PPC issue will require some of the PPC* guys to look.  This (PPC*) has been broke for a LONG time.

masatake:
... I think we shuold not make a release now. I will try to fix #67.

masatake:
> This (PPC*) has been broke for a LONG time.
Do you mean lsof is broken? or the kernel header is broken?

If lsof is broken, it sill better to write it to somewhere our documents.

masatake:
I found the following code in sys/refcount.h:
```
#ifndef __SYS_REFCOUNT_H__
#define __SYS_REFCOUNT_H__

#include <sys/limits.h>
#include <machine/atomic.h>

#ifdef _KERNEL
#include <sys/systm.h>
#else
#include <stdbool.h>
#define KASSERT(exp, msg)       /* */
#endif
```

So I wonder why we must include stdbool.h or sys/param.h explictly.

masatake:
contrib/ipfilter/ipsend/sock.c:
```
#if defined(__NetBSD__) && defined(__vax__)
/*
 * XXX need to declare boolean_t for _KERNEL <sys/files.h>
 * which ends up including <sys/device.h> for vax.  See PR#32907
 * for further details.
 */
typedef int     boolean_t;
#endif
```

lrosenman:
Per a note from kib@FreeBSD.org (on the freebsd-current@FreeBSD.org mailing list), it's needed as the minimum fix for the change  post r350199. 

I haven't chased all the headers, and dlsof.h is an #ifdef nightmare.

masatake:
I this I should not say much because I don't use FreeBSD daily.
However, including stdbool.h on lsof side is a bit strange for me because lsof itself doesn't use bool.

sys/types.h has following lines.
```

#if !defined(__bool_true_false_are_defined) && !defined(__cplusplus)
#define	__bool_true_false_are_defined	1
#define	false	0
#define	true	1
#if __STDC_VERSION__ < 199901L && __GNUC__ < 3 && !defined(__INTEL_COMPILER)
typedef	int	_Bool;
#endif
typedef	_Bool	bool;
#endif /* !__bool_true_false_are_defined && !__cplusplus */
```

It seems that bool is defined in sys/types.h. Can't we use the definition by including sys/types.h.

thexcoder:
Hi, tested this and posted the reply in #69

thexcoder:
Hi, tested this and posted the reply in #69

lrosenman:
I'm following what the FreeBSD kernel developer who added the requirement is saying.  He says to include <stdbool.h>. 

I really wish some of the folks who joined this project from FreeBSD would help here.  I'm not the best C programmer but want to see lsof continue to support FreeBSD.

masatake:
Can I access the archive of the discssion on the mailing list?

masatake:
I find another interesting code in lsof side:
```
#if     defined(NEEDS_BOOL_TYPEDEF)
/*
 * Under FreeBSD 10.0 on the i386 architecture the bool typedef is not defined
 * unless <sys/types.h> is #included under _KERNEL.  As explained with
 * cpumask_t, #include'ing <sys/types.h> under _KERNEL causes other problems.
 * Hence Configure specifies that bool be typedef'd here specifically for the
 * i386 architecture on FreeBSD 10.0.
 */
# if    __STDC_VERSION__<199901L && __GNUC__<3 && !defined(__INTEL_COMPILER)
typedef int     _Bool;
# endif /*__STDC_VERSION__<199901L && __GNUC__<3 && !defined(__INTEL_COMPILER)*/
typedef _Bool   bool;
#endif  /* defined(NEEDS_BOOL_TYPEDEF) */
```

lrosenman:
thread on freebsd-current:
https://lists.freebsd.org/pipermail/freebsd-current/2019-July/073901.html

masatake:
@lrosenman, thank you. I read the thread.
It seems that you follow the comment in https://lists.freebsd.org/pipermail/freebsd-current/2019-July/073908.html:
```
Absolutely minimal change to accomodate updated refcount.h would be
inclusion of stdbool.h.
```

However, I think this is not a suitable idea for us in the current situation. What we need is a something maintainable fix. So rather I think we should follow the comment in https://lists.freebsd.org/pipermail/freebsd-current/2019-July/073904.html:
```
Then you have to follow the rules (from style(9))...

   If either <sys/types.h> or <sys/param.h> is needed, include it
   before other include files.  (<sys/param.h> includes <sys/types.h>;
   do not include both.)

Typically, if you're including anything from sys/ you'll need
sys/types.h first.
```
dialect/freebsd/machine.h includes sys/types.h. It implies Vic knows the rules.
The machine.h is included by lsof.h. So if a .c file includes lsof.h, we can avoid the trouble.
dialects/freebsd/{dmnt.c,dnode1.c,dnode.c,dproc.c,dsock.c,dstore.c}, all freebsd .c files other than dnode2.c include lsof.h at the head. The exception is dnode2.c.

So I GUESS the maintainable fix is:
```
diff --git a/dialects/freebsd/dnode2.c b/dialects/freebsd/dnode2.c
index d1c6e0d..af05414 100644
--- a/dialects/freebsd/dnode2.c
+++ b/dialects/freebsd/dnode2.c
@@ -38,6 +38,8 @@ static char copyright[] =
 #endif
 
 
+#include "lsof.h"
+
 #if	defined(HAS_ZFS)
 
 #define _KERNEL
``` 

It seems that Vic handled dnode2.c very specially. dnode2.c has its own makefile (Makefile.zfs).
So I cannot convince myself including lsof.h solves the issue. 

Could you try my change instead of including stdbool.h directly.

lrosenman:
Given that:
1) this is a FreeBSD specific file
2) Vic did *NOT* include lsof.h in dnode2.c
3) I've already patched the port to include <stdbool.h>

I really don't think that your change is appropriate.
Until/unless we get some help from the FreeBSD folks that joined that have the actual C programming skills and Kernel knowledge to help unwind the ifdef mess, I'd like to just commit this fix (including stdbool.h) to master.

masatake:
> this is a FreeBSD specific file

Which file?

> Vic did NOT include lsof.h in dnode2.c
Really? I think Vic forgot including lsof.h in dnode2.c.
Till now, the forgetting has not become an issue. 

I think https://lists.freebsd.org/pipermail/freebsd-current/2019-July/073904.html shows the very fundamental  rule that applications including lsof should follow. 
People at the mailing list already helps us enough.

masatake:
It seems that I was wrong.
> Vic did NOT include lsof.h in dnode2.c
I agree wit this. What I wrote may be wrong.

Vid did not include lsof.h in dnode2.c intentionally.

zfs may be very special.

Vic cannot include lsof.h because zfs is so special.
Instead, he introduced dzfs.h for handling extra data types.

If the issue is related only to dnode2.c, I think we have to fix dzfs.h.

ckkv.c is failed to compile. It looks strange.
ckkv.c includes machine.h directly.
machine.h includes 
```
#include <sys/types.h>
#include <sys/param.h>
```
. This follows the rule shown in  https://lists.freebsd.org/pipermail/freebsd-current/2019-July/073904.html  .

lrosenman:
with adding <stdbool.h> to dlsof2.c, ckkv.c compiles.

see:
http://home.lerctr.org:8888/data/live-host-ports/2019-07-26_10h08m21s/logs/lsof-4.93.2_3,8.log

This is with the *EXACT* fix I propose here.

masatake:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=240349

masatake:
@lrosenman, I'm very sorry to be late to take action for this issue.
For a year, I have tried to find time for studying FreeBSD. However, I cannot find enough time.

Not I understand I should say nothing about dialects I don't know well as far as there is a person who wants to maintain the dialects.

If this change is still correct and needed on FreeBSD, please, merge this.

Here after I will not review the code for other dialects than linux. Please, do as you want about the code of FreeBSD dialect.

However, I will take care of common files.

PR #73, opened 2019-10-28
Add missing header to FreeBSD dlsof.h

william-gr:
Without this header the build fails on latest 12-STABLE and possibly 13-CURRENT.


lrosenman:
Approved

PR #83, opened 2019-12-09
.gitignore: Ignore top-level Makefile only

iv-m:
Without '/', all the Makefiles are ignored, including
dialect-specific ones (unless they are already added),
which can be hard-to-pin source of confusion.


masatake:
Good catch!

masatake:
Thank you!

PR #94, opened 2020-06-04
[linux] Allow reproducible builds

T4cC0re:
In a reproducible build all varied information is removed.
This patch does so, by checking if the standard [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/) variable is set.
If it is, we are attempting a reproducible build and will strip varying information.

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -337,7 +337,9 @@ provided test systems where I was able to do development work.
 	Andrew Merril
 	Richard van Meurs
 	Jim Mewes
+	Ivan Melnikov
 	Conrad Meyer
+	Hendrik Meyer
 	Gary Millen
 	Timothy Miller
 	Davin Milun
@@ -538,8 +540,7 @@ provided test systems where I was able to do development work.
 	Donald Zoch
 	Malcom Zung
 	Waldemar Zurowski
-	@eranik (github account)
-	Ivan Melnikov and
+	@eranik (github account) and
 	@jolmg  (github account)
 
 If I have omitted a contributor's name, the fault is wholly mine,
--- a/00DIST
+++ b/00DIST
@@ -5039,5 +5039,13 @@ July 14, 2018
 		in command line doesn't exist.
 		This closes the github issue #90 reported by @rowlap.
 
+		[linux] Allow reproducible builds
+		In a reproducible build all varied information is removed.  This
+		change does so, by checking if the standard SOURCE_DATE_EPOCH
+		variable is set.  If it is, we are attempting a reproducible build
+		and will strip varying information.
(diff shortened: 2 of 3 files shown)


eli-schwartz:
Thanks for working on this! Note that for reproducible builds we need to avoid embedding the time zone also, so that date call (for the SOURCE_DATE_EPOCH case) should also use `-u`.

See also "reading the variable": https://reproducible-builds.org/docs/source-date-epoch/

T4cC0re:
@eli-schwartz Thanks for noticing. I just pushed an amended commit with the fix.

masatake:
@ lrosenman, I will merge this though testing on freebsd is failed because the change is nothing to do with other dialect than Linux.

lrosenman:
@masatake the FreeBSD CI is failed due to the Cirrus stuff using outdated images. 
I'll see if I can figure out how to fix that for all 3 FreeBSD versions.

masatake:
@lrosenman, thank you.

T4cC0re:
@masatake Is there a procedure to updating the 00DIST file? Or do I just edit it by hand?

masatake:
@T4cC0re, please, edit the file by hand directly.

T4cC0re:
@masatake There is a Footnote in the bootom of the file with your name and a date. Should I change the date or leave it be?

masatake:
Please, keep my name as is. Please, write what you did as an item like:
```

		[linux] fix a crash when printing the endpoint for unaccepted
		unix socket with +E option.
		This closes the github issue #74 reported by @jolmg.
```
The number of pull request (#94) and your name should be written in the item.

Updating 00CREDITS is ... my job.

T4cC0re:
@masatake I updated the 00DIST file, does that look good to you?

masatake:
@T4cC0re, yes thank you. I will merge this, and we will make a release in ... a month.

lrosenman:
The FreeBSD CI images are changed in master now, but until #97 is merged they will fail.  Ths should be fine to merge.

masatake:
@lrosenman, though you merged #97, the tests of this pull request are failed.
Are they expected?

lrosenman:
Yes, until I can get the kernel folks to look at the code and fix getting the executable and library names.

masatake:
@lrosenman, I see. Thank you.
After updating 00CREDITS, I will merge this.

masatake:
@T4cC0re, thank you.

bmwiedemann:
In openSUSE, we had
https://github.com/bmwiedemann/openSUSE/blob/master/packages/l/lsof/lsof-no-build-date-etc.patch from 2010 for reproducible-builds that will soon not be needed anymore, thanks to this patch.

PR #95, opened 2020-06-05
update FreeBSD images

lrosenman:
update to modern FreeBSD images


lrosenman:
These CI checks WILL FAIL until I merge all the FreeBSD local patches I have.  BUT it has the new images (most current as of today). 

@masatake can I merge this to make the other merges easier?

And should I pick up ALL the local patches I have in one PR or multiple PR's ?

masatake:
@lrosenman, I would like you to as you want. This has not impact on the other dialects and the change is user-invisble.  So, please, do as your comfortable way.

PR #99, opened 2020-06-10
Fix tests for large inode-numbers (i >= 2^32).

hvpeteet:
The way that test strings were formatted could lead to tests failing
when run with large inode-numbers (i >= 2^32). This is fixed by
formatting with uint64 values intead of unsigned ints.

### DIFF
--- a/tests/LTbasic.c
+++ b/tests/LTbasic.c
@@ -356,8 +356,8 @@ tstlsof(texec, tkmem, tproc)
 		    pem = cem;
 		    break;
 		}
-		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%u",
-		    (unsigned int)cwdsb.st_ino);
+		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%" PRIu64,
+		    (uint64_t)cwdsb.st_ino);
 		ibuf[sizeof(ibuf) - 1] = '\0';
 		if ((tmpdc.maj == cwddc.maj)
 		&&  (tmpdc.min == cwddc.min)
@@ -389,8 +389,8 @@ tstlsof(texec, tkmem, tproc)
 		    pem = cem;
 		    break;
 		}
-		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%u",
-		    (unsigned int)kmemsb.st_ino);
+		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%" PRIu64,
+		    (uint64_t)kmemsb.st_ino);
 		ibuf[sizeof(ibuf) - 1] = '\0';
 		if ((tmpdc.maj == kmemdc.maj)
 		&&  (tmpdc.min == kmemdc.min)
@@ -420,8 +420,8 @@ tstlsof(texec, tkmem, tproc)
 		    pem = cem;
 		    break;
 		}
-		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%u",
-		    (unsigned int)lsofsb.st_ino);
+		(void) snprintf(ibuf, sizeof(ibuf) - 1, "%" PRIu64,
+		    (uint64_t)lsofsb.st_ino);
 		ibuf[sizeof(ibuf) - 1] = '\0';
 		if ((tmpdc.maj == lsofdc.maj)
 		&&  (tmpdc.min == lsofdc.min)
--- a/tests/LTdnlc.c
+++ b/tests/LTdnlc.c
@@ -186,7 +186,7 @@ main(argc, argv)
     }
     if ((em = ConvStatDev(&sb.st_dev, &cwddc)))
 	PrtMsgX(em, Pn, cleanup, 1);
-    (void) snprintf(ibuf, sizeof(ibuf) - 1, "%u", (unsigned int)sb.st_ino);
+    (void) snprintf(ibuf, sizeof(ibuf) - 1, "%" PRIu64, (uint64_t)sb.st_ino);
     ibuf[sizeof(ibuf) - 1] = '\0';
 /*
  * Loop ATTEMPT_CT times.
--- a/tests/LTnlink.c
+++ b/tests/LTnlink.c
@@ -209,7 +209,7 @@ print_file_error:
  */
     if ((em = ConvStatDev(&tfsb.st_dev, &tfdc)))
 	PrtMsgX(em, Pn, cleanup, 1);
-    (void) snprintf(ibuf, sizeof(ibuf) - 1, "%u", (unsigned int)tfsb.st_ino);
+    (void) snprintf(ibuf, sizeof(ibuf) - 1, "%" PRIu64, (uint64_t)tfsb.st_ino);
     ibuf[sizeof(szbuf) - 1] = '\0';
     (void) snprintf(xlnk, sizeof(xlnk) - 1, "%d", (int)tfsb.st_nlink);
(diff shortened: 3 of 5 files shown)


masatake:
@lrosenman, could you try this one on FreeBSD?

lrosenman:
Given that LTBasic is *ALREADY* failing on FreeBSD, I can't really test it and suspect it'll be fine, as we've had 64-bit Inodes for a bit. 

When/if we get the FreeBSD stuff to find libraries/executables again, I'll see what happens, and propose fixes if needed.

masatake:
I see. Thank you.

masatake:
@hvpeteet, do you think we have to fix not only the test harness but also the lsof itself about the inode size?

```
[yamato@slave]~/var/lsof% grep -r st_ino dialects/linux 
dialects/linux/dsock.c:		    up->sb_ino = (INODETYPE)sb.st_ino;
dialects/linux/dsock.c:	&&  (ap = check_ax25((INODETYPE)s->st_ino))
dialects/linux/dsock.c:	&&  (ip = check_ipx((INODETYPE)s->st_ino))
dialects/linux/dsock.c:		Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	&&  (rp = check_raw((INODETYPE)s->st_ino))
dialects/linux/dsock.c:		Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	    &&  (np = check_netlink((INODETYPE)s->st_ino))
dialects/linux/dsock.c:	    Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	&&  (pp = check_pack((INODETYPE)s->st_ino))
dialects/linux/dsock.c:		    (INODETYPE)s->st_ino);
dialects/linux/dsock.c:	&&  (up = check_unix((INODETYPE)s->st_ino))
dialects/linux/dsock.c:	    Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	&&  (rp = check_raw6((INODETYPE)s->st_ino))
dialects/linux/dsock.c:		Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	&&  (tp6 = check_tcpudp6((INODETYPE)s->st_ino, &pr))
dialects/linux/dsock.c:		    (INODETYPE)s->st_ino);
dialects/linux/dsock.c:		Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	&&  (tp = check_tcpudp((INODETYPE)s->st_ino, &pr))
dialects/linux/dsock.c:		    (INODETYPE)s->st_ino);
dialects/linux/dsock.c:		Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	if ((ss & SB_INO) && (sp = check_sctp((INODETYPE)s->st_ino))
dialects/linux/dsock.c:	    (void) snpf(tbuf, sizeof(tbuf), InodeFmt_d, (INODETYPE)s->st_ino);
dialects/linux/dsock.c:	&&  (icmpp = check_icmp((INODETYPE)s->st_ino))
dialects/linux/dsock.c:	    Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dsock.c:	    Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dlsof.h:#define	SB_INO		0x02		/* st_ino */
dialects/linux/dlsof.h:	INODETYPE inode;		/* directory st_ino */
dialects/linux/dmnt.c:	    mp->inode = (INODETYPE)((ds & SB_INO) ? sb.st_ino : 0);
dialects/linux/dnode.c:	    Lf->inode = (INODETYPE)s->st_ino;
dialects/linux/dproc.c:	    &&  (dev == s->st_dev) && (inode == (INODETYPE)s->st_ino))
dialects/linux/dproc.c:		sb.st_ino = (ino_t)inode;
dialects/linux/dproc.c:	    } else if ((sb.st_dev != dev) || ((INODETYPE)sb.st_ino != inode)) {
dialects/linux/dproc.c:			    ((INODETYPE)sb.st_ino == inode) ? ")" : ",");
dialects/linux/dproc.c:		    if ((INODETYPE)sb.st_ino != inode) {
dialects/linux/dproc.c:			    sep, (INODETYPE)sb.st_ino);
dialects/linux/dproc.c:		sb.st_ino = (ino_t)inode;
dialects/linux/dproc.c:		Lf->inode = (ino_t)sb.st_ino;
```
```
[yamato@slave]~/var/lsof% grep -r INODETYPE *.h dialects/linux/*.h
dlsof.h:	INODETYPE inode;		/* directory st_ino */
dlsof.h:	INODETYPE i;			/* inode number */
dlsof.h:	INODETYPE ino;			/* file's inode */
lsof.h:# if	!defined(INODETYPE)
lsof.h:#define	INODETYPE	unsigned long	/* node number storage type */
lsof.h:# endif	/* !defined(INODETYPE) */
lsof.h:	INODETYPE inode;		/* inode number */
lsof.h:	INODETYPE inode;
lsof.h:	INODETYPE inode;
lsof.h:	INODETYPE fs_ino;		/* file system inode number */
lsof.h:	INODETYPE inode;		/* search inode number */
machine.h: * INODETYPE and INODEPSPEC define the internal node number type and its
machine.h:#define	INODETYPE	unsigned long long
machine.h:#define	INODEPSPEC	"ll"	 	/* INODETYPE printf specification
dialects/linux/dlsof.h:	INODETYPE inode;		/* directory st_ino */
dialects/linux/dlsof.h:	INODETYPE i;			/* inode number */
dialects/linux/dlsof.h:	INODETYPE ino;			/* file's inode */
dialects/linux/machine.h: * INODETYPE and INODEPSPEC define the internal node number type and its
dialects/linux/machine.h:#define	INODETYPE	unsigned long long
dialects/linux/machine.h:#define	INODEPSPEC	"ll"	 	/* INODETYPE printf specification
```

hvpeteet:
Thanks for the pointers I wasn't really familiar with the codebase but now I see the real issue was the test not using INODEPSPEC and INODETYPE (instead it used an unsigned int). 

Going forward I think the fix is to replace
1. PRIu64 --> INODEPSPEC "u"
1. uint64_t --> INODETYPE
1. unsigned long long --> uint64_t (**optional** based on your reply)

Would you prefer I include the optional bit or leave it as it is?

masatake:
Sorry to be late. I didn't notice your reply. I will take a look into your comment.

hvpeteet:
I think that what I described above would be ideal, but just using this patch would be a good step forward (as it would stop breaking the tests). The reason I am wary about trying to implement the whole comment from above is that I don't know how easy it is to inject the correct headers based on a dialect into the tests. Since each dialect can redefine those macros it may just be easier to use the uint64_t in the tests and leave the headers as is.

hvpeteet:
poke for update?

masatake:
I'm very sorry to be late.

> 4. 'll' (for INODEPSPEC) --> PRIu64 (optional based on your reply) 
>    o For this reason I don't think we should change the optional bit since PRIu64 includes the "u" character and INODESPEC does not. If we changed it here we would need to change the other defs for it as well as main.c. IMO just the first 2 changes are good enough to fix the bug and keep the code consistent.

INODESPEC doesn't include 'u' but 'u' is appended to the format string where INODESPEC is used.
So I will not change the definition of INODETYPE and INODESPEC.
I will consider 1. and 2. after merging this.

PR #107, opened 2020-09-27
Improve man page

kurahaupo:
The "piped to" description is flat wrong, because "kill" doesn't read anything from stdin.
Saying that one command line option "selects" another is weird; "implies" would be clearer, both times it's mentioned.

Also added some examples.

### DIFF
--- a/Lsof.8
+++ b/Lsof.8
@@ -1692,8 +1692,9 @@ header), so that it is easy to use programmatically. e.g.
 	done
 
 .fi
+The
 .B \-t
-selects the
+option implies the
 .B \-w
 option.
 .TP \w'names'u+4
@@ -1778,7 +1779,7 @@ them when already enabled is acceptable.
 .IP
 The
 .B \-t
-option selects the
+option implies the
 .B \-w
 option.
 .TP \w'names'u+4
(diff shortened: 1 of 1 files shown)


Lsof.8
```diff
@@ -1676,11 +1676,20 @@ FAQ (The \fBFAQ\fP section gives its location.)
...
-e.g., so that the output may be piped to
-.IR xargs kill.
+produce terse output comprising only process identifiers (without a
+header), so that it is easy to use programmatically. e.g.
+.nf
+
+	# reload anything using old SSL
+	lsof -t /lib/*/libssl.so.* | xargs -r kill -HUP
+
+        # get list of processes and then iterate over them
```
masatake: It seems that one extra white space is at the start of the line. Could you remove it for aligning.

masatake:
Thank you.

Please, unify the two commits, b5acab5 and 7e8a2fa into one commit.

Could you use "man page: " as the prefix for the commits?
This helps me to update 00DIST file.

masatake:
Do you allow me to rearrange the commits?

I guess you are using GitHub's web UI. It may be hard to reflect my comments to your change with the UI.

kurahaupo:
Done. (And properly attributed the commits to my normal public persona.)

masatake:
Thank you.

masatake:
Putting examples to the man page is a good idea.

Issue #102, opened 2020-08-17
Ipv6 is long address, Command lsof  can't display  port to the log address


zhrf2020:
Repetition scenario：
1  Add ipv6
ip -6 addr add abcd:ef10:ffff:ffff:ffff:ffff:ffff:ff62 dev enp1s0
ip -6 addr add abcd:ef10::ff62 dev enp1s0   
 
2  The test code （python2.7）
code1:
#vim test_log.py
import socket,sys,time 
tcpT6Server=socket.socket(socket.AF_INET6,socket.SOCK_STREAM)
tcpT6Server.bind(("abcd:ef10:ffff:ffff:ffff:ffff:ffff:ff62",8888))
tcpT6Server.listen(5)
time.sleep(10000)

code2: 
#vim test.py
import socket,sys,time 
tcpT6Server=socket.socket(socket.AF_INET6,socket.SOCK_STREAM)
tcpT6Server.bind(("abcd:ef10::ff62",8888))
tcpT6Server.listen(5)
time.sleep(10000)

3 Run  codes 
#python  test_log.py
#python  test.py

4 Execution port command
#lsof -i:8888
#netstat -anpW | grep 8888

5  Results
Command lsof can't display ipv6 address (abcd:ef10:ffff:ffff:ffff:ffff:ffff:ff62) port number 8888

6   Code snippet
 * Assemble four uint32_t's from 4 X 8 hex digits into s6_addr32[].
 */
	for (i = 0, len = strlen(as);
	     (i < 4) && (len >= 8);
	     as += 8, i++, len -= 8)
	{
	    (void) strncpy(buf, as, 8);
	    buf[8] = '\0';
	    ep = (char *)NULL;
	    if ((ad->s6_addr32[i] = (uint32_t)strtoul(buf, &ep, 16))
	    ==  (uint32_t)UINT32_MAX || !ep || *ep)
		break;
	}
	return((*as || (i != 4) || len) ? 1 : 0);
}
#endif	/* defined(HASIPv6) */

There seems to be wrong ：
            if ((ad->s6_addr32[i] = (uint32_t)strtoul(buf, &ep, 16))
	     ==  (uint32_t)UINT32_MAX || !ep || *ep)

Correct code:
           if (   (ad->s6_addr32[i] = (uint32_t)strtoul(buf, &ep, 16))
	     == ((uint32_t)UINT32_MAX || !ep || *ep)   )

masatake:
Why don't you make a pull request for this?
[...]

PR #109, opened 2020-09-29
Fix an issue that ipv6 is a special form,the information cannot be displayed

zhrf2020:
Fix an issue that ipv6 is a special form,the information cannot be displayed

### DIFF
--- a/dialects/linux/dsock.c
+++ b/dialects/linux/dsock.c
@@ -3633,7 +3633,7 @@ net6a2in6(as, ad)
 	    buf[8] = '\0';
 	    ep = (char *)NULL;
 	    if ((ad->s6_addr32[i] = (uint32_t)strtoul(buf, &ep, 16))
-	    ==  (uint32_t)UINT32_MAX || !ep || *ep)
+	    >  (uint32_t)UINT32_MAX || !ep || *ep)
 		break;
 	}
 	return((*as || (i != 4) || len) ? 1 : 0);


masatake:
Thank you for making this pull request.

Could you put the original report (and analysis) that you wrote in the initial comment of #102 to the commit log?
In addition, please put "close #102" at the beginning of the commit log.

This is a Linux dialect-specific fix. So, put `[linux]` as a prefix for the commit log.

The commit log will be like:
```
[linux] fix an issue that the information cannot be displayed if ipv6 addr is a special form

Close #102.

The bug fixed by this change:
Command lsof can't display port to the long address if Ipv6 is long address.

Repetition scenario：
1 Add ipv6
ip -6 addr add abcd:ef10:ffff:ffff:ffff:ffff:ffff:ff62 dev enp1s0
ip -6 addr add abcd:ef10::ff62 dev enp1s0
...
```
(I'm not good at English. So, please, change the title and body as you want .)

dialects/linux/dsock.c
```diff
@@ -3633,7 +3633,7 @@ net6a2in6(as, ad)
 	    buf[8] = '\0';
 	    ep = (char *)NULL;
 	    if ((ad->s6_addr32[i] = (uint32_t)strtoul(buf, &ep, 16))
-	    ==  (uint32_t)UINT32_MAX || !ep || *ep)
+	    >  (uint32_t)UINT32_MAX || !ep || *ep)
```
masatake, line 3636: In my understanding, the value (V) of the left side of `>` operator is never greater than (uint32_t)UINT32_MAX.

masatake:
I reviewed your code carefully. See my inline comment.
I think we can remove the condition `>  (uint32_t)UINT32_MAX`. However, I would like to propose more redandant code for fixing #102. I will make a pull request. I would like you to review it.

PR #119, opened 2020-10-14
Make -Fo option work

### LINKED ISSUE #118: Behaviour of offset field on -F inconsistent with the man page
Using lsof 4.91 and 4.93.2 at least (Debian stable and sid), the offset field for `-F` does not work according to the man page documentation (or my understanding of it).

When `FILE` is opened by exactly one process, I'd expect `lsof -Fo FILE` to produce three lines similar to:

```
p676080
f4
o49152
```

Instead, it only prints the first two lines, i.e. the offset is omitted. When `-F` isn't used, the offset has to be enabled explicitly with `-o` (at least for regular files). This works, but the output is:

```
p676080
f4
o0t49152
```

The `0t` prefix is used on the normal output (i.e. without `-F`) to distinguish between decimal and hexadecimal notation. But the `o` field on `-F` is always decimal, so I'd expect it to not include that prefix. This would match the `s` field, which is also defined to be decimal and does not include the prefix. Also, the man page explicitly mentions the `0x` prefix for the hexadecimal fields (e.g. `D`, `F`, `r`) but does not include `0t` for `o`.

So the two changes I'd suggest are enabling `-o` when the `-F` list includes `o` and removing the `0t` prefix from the field in that output mode.

masatake:
Close #118.
    
-Fo option is for printing file offset. For regular files, the option didn't work.
Here is a command session demonstrating the fix:
    
        # ./lsof -Fo -o0| grep ^o | sort | uniq -c
        90586 o0t0
           87 o0t101
           84 o0t103
        ...
    
 @JustAnotherArchivist reported this bug in #118.

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -547,8 +547,9 @@ provided test systems where I was able to do development work.
 	jamie@catflap.org (email)
 	@hardikpnsp (github account)
 	Martin D Kealey
-	Henry Peteet and
-	@zhrf2020
+	Henry Peteet
+	@zhrf2020 and
+	@JustAnotherArchivist
 
 If I have omitted a contributor's name, the fault is wholly mine,
 and I apologize for the error.
--- a/00DIST
+++ b/00DIST
@@ -5139,5 +5139,19 @@ July 14, 2018
 		man page,[linux]: enumerate abbreviated flags printed with '+f g' option
 
 
+		Make -Fo option work
+		-Fo option is for printing file offset. For regular files,
+		the option didn't work.
+
+		Here is a command session demonstrating the fix:
+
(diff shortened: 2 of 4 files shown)


masatake:
This project started in 4.92. https://github.com/lsof-org/lsof/blob/master/00DIST#L4915
We don't know much about how the original author had developed lsof.

I, a former maintainer, don't know much about macOS. However, it seems that lsof 4.99.5 for macOS is available at https://formulae.brew.sh/formula/lsof. Isn't it enough?

forthrin:
@masatake: Just that bundling would make for one less dependency, but sure am thankful for its continued development that made the creation of a progress meter into a piece of cake with the readily available fseek for file handlers <3

PR #121, opened 2020-10-15
[freebsd] cirrus-ci: disabled

masatake:
https://github.com/lsof-org/lsof/pull/94#issuecomment-639180877h

    lrosenman commented on Jun 5

    @masatake the FreeBSD CI is failed due to the Cirrus stuff using outdated images.
    I'll see if I can figure out how to fix that for all 3 FreeBSD versions.

Till we find the way to update the images, disable the testing.


masatake:
@lrosenman, revert this change when the images for FreeBSD are updated.

PR #123, opened 2020-10-15
Fix man page definition of the `o` field on programmatic output

JustAnotherArchivist:
#118


masatake:
Thank you. I will merge this in soon.

masatake:
Thank you.

PR #124, opened 2020-10-17
Pidfd pid

masatake:


masatake:
* NSPid should be captured, too.
* A test case is needed.

PR #125, opened 2020-10-20
Don't select the file descriptor field by default.

### LINKED ISSUE #103: lsof -F p prints more than just pid
In earlier versions of lsof, -F p printed only the pid:

$ sudo lsof -i tcp:22 -s tcp:LISTEN -F p
p677
$ rpm -q lsof
lsof-4.87-6.el7.x86_64

Newer versions print filehandles also, which breaks scripts parsing this output:

$ sudo lsof -i tcp:22 -s tcp:LISTEN -F p
p609
f5
f6
$ rpm -q lsof
lsof-4.93.2-3.fc32.x86_64

Is this behavior change intentional?

masatake:
Close #103.

The version 4.88 introduced the change for selecting the file
descriptor field by default. However, the change is not
suitable for users who wants to print only PID field.
@po5857 suggests the use case.

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -548,8 +548,9 @@ provided test systems where I was able to do development work.
 	@hardikpnsp (github account)
 	Martin D Kealey
 	Henry Peteet
-	@zhrf2020 and
-	@JustAnotherArchivist
+	@zhrf2020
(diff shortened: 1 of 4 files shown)


masatake:
A test case for the change is needed.

PR #144, opened 2020-11-23
tests: Don't ignore failures

nckx:
GNU Guix makes the ‘standard’ and ‘optional’ tests and expects both to
return non-zero exit status on failure. However, only ‘make silent’ did.
This led to test failures going unnoticed.

Make all targets properly signal failure (and even more obvious how much
boilerplate they share :-).

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -27,7 +27,7 @@ all:	${CKTSTDB} ${BASTST} ${STDTST} FRC
 	  exit 1 ;\
 	fi
 	@rm -f config.LT*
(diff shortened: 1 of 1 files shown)


masatake:
Thank you.

masatake:
However, we don't use Makefile at all in the CI/CD environment.
We use "make check" at the top level of the source tree.

nckx:
Thanks for merging!

> We use "make check" at the top level of the source tree.

...and for the advice.  I'll try to change our package to do the same.

PR #150, opened 2021-01-26
Drop LSOF_CCDATE in order to ensure reproducible builds

dilinger:
LSOF_CCDATE is used to embed the build time/date string into the lsof binary.
This string is displayed as part of 'lsof -v' output. However, having that
embedded string breaks reproducible builds - we want to ensure that subsequent
builds (using the same exact build environemnt) produce the same exact binary,
bit-for-bit.

Debian has a patch that allows overriding the variable, similar to what lsof allows
for LSOF_HOST (and others). Let me know if you'd prefer that patch. However,
I'm of the opinion that LSOF_CCDATE should just be dropped completely. We
can see the version to tell us what the corresponding source code is, and we have
file timestamps to tell us when a binary was installed. LSOF_CCDATE provides
no security, as an attacker could hardcode the older date/time string in their
malicious binary. So, this pull request just drops the variable completely.

### DIFF
--- a/00DIST
+++ b/00DIST
@@ -5229,6 +5229,15 @@ July 14, 2018
 		Provided by @dilinger (Andres Salomon) in #149.
 		A commit in the pull request includes work of Nicholas Bamber.
 
+		Drop LSOF_CCDATE across all dialects to ensure reproducible builds
+		Simplify things for reproducible builds by just getting rid of
+		the embedded date/time string. With LSOF_CCDATE gone, keeping
+		SOURCE_DATE_EPOCH around doesn't make much sense, so drop that as
+		well. Folks doing reproducible builds should still override the
+		LSOF_HOST, LSOF_LOGNAME, LSOF_SYSINFO, and LSOF_USER variables (as
+		they were previously doing before SOURCE_DATE_EPOCH).
+		Provided by @dilinger (Andres Salomon) in #150.
+
 
 The lsof-org team at GitHub
 November 11, 2020
--- a/NEW/usage.c
+++ b/NEW/usage.c
@@ -710,8 +710,6 @@ usage(xv, fh, version)
 		(void) fprintf(stderr, "    configuration info: %s\n", cp);
 #endif	/* defined(LSOF_CINFO) */
 
-	    if ((cp = isnullstr(LSOF_CCDATE)))
-		(void) fprintf(stderr, "    constructed: %s\n", cp);
 	    cp = isnullstr(LSOF_HOST);
 	    if (!(cp1 = isnullstr(LSOF_LOGNAME)))
 		cp1 = isnullstr(LSOF_USER);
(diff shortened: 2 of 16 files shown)


dilinger:
By the way - there are additional variables that seem completely unnecessary that I could also remove if you'd like. On my system after I build, I see the following output from 'lsof -v':

lsof version information:
    revision: 4.93.2
    latest revision: https://github.com/lsof-org/lsof
    latest FAQ: https://github.com/lsof-org/lsof/blob/master/00FAQ
    latest (non-formatted) man page: https://github.com/lsof-org/lsof/blob/master/Lsof.8
    constructed: Tue 26 Jan 2021 07:04:04 AM UTC
    constructed by and on: root@e7470
    compiler: cc
    compiler version: 10.2.1 20210110 (Debian 10.2.1-6) 
    compiler flags: -g -O2 -ffile-prefix-map=/x/lsof-4.93.2+dfsg=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -DLINUXV=510009 -DGLIBCV=231 -DHASIPv6 -DNEEDS_NETINET_TCPH -DHASSELINUX -DHASUXSOCKEPT -DHASPTYEPT -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -DHAS_STRFTIME -DLSOF_VSTR="5.10.9" -O
    loader flags: -L./lib -llsof -Wl,-z,relro -Wl,-z,now -lselinux
    system info: Linux e7470 5.10.0-1-amd64 #1 SMP Debian 5.10.4-1 (2020-12-31) x86_64 GNU/Linux
    Anyone can list all files.
    /dev warnings are disabled.
    Kernel ID check is disabled.

Distributions often override those variables (Debian overrides LSOF_CCDATE, LSOF_HOST, LSOF_LOGNAME, LSOF_SYSINFO, and LSOF_USER).

I'm not sure who actually cares what compiler was used (LSOF_CC), or when it was built (LSOF_CCDATE) and by who (LSOF_USER/LSOF_LOGNAME) and where (LSOF_HOST). I'm happy to add additional commits to this branch to remove more of those variables if you agree with removing any of them. LSOF_SYSINFO in particular seems very unneccessary.

masatake:
We have already an applied patch for the same purpose.
https://github.com/lsof-org/lsof/pull/94/files
How do you think about this?

masatake:
Dropping the coverage is not important in this pull request, of course.

dilinger:
> We have already an applied patch for the same purpose.
> https://github.com/lsof-org/lsof/pull/94/files
> How do you think about this?

That will also work. The Debian patch (which you can see here: 
https://sources.debian.org/src/lsof/4.93.2+dfsg-1.1/debian/patches/preset-ccdate/ ) changes it for all dialects. For Debian, the only ones we actually care about are Linux and FreeBSD. I can submit a separate PR for freebsd if you'd like.

Personally I'd still rather see a bunch of the noise removed, but it's your choice.

masatake:
Could you consider writing an entry for your change at the end of 00DIST file?

dilinger:
Sure, but how should I be updating this branch? Removing CCDATE or adding the freebsd override using SOURCE_DATE_EPOCH?

masatake:
Let's remove CCDATE.

masatake:
Thank you. I will review (and merge).

PR #151, opened 2021-03-07
FreeBSD isofs fixes and msdosfs support

DamjanJovanovic:
Fixes for the ISO9660 filesystem. Adds support for msdosfs.

### DIFF
--- a/00DIST
+++ b/00DIST
@@ -5246,5 +5246,8 @@ July 14, 2018
 		again. Fix the header inclusion order. Also add the new way of finding
 		dev_t on more recent FreeBSD versions.
 
+
+		[FreeBSD] add support for msdosfs on FreeBSD
+
 The lsof-org team at GitHub
 November 11, 2020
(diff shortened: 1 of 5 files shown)


masatake:
I'm the maintainer of linux dialect. @lrosenman, the maintainer of FreeBSD dialect may review your changes.

Dialect independent comments:

Could you add "freebsd: " as the prefix for the commit headers like:
```
freebsd: the ISO9660 filesystem broke starting with FreeBSD 7 due to the header
freebsd: add support
```

In addition, could you add entries at the end of [00DIST](https://github.com/lsof-org/lsof/blob/master/00DIST)?

DamjanJovanovic:
Thank you, I've changed the commit messages and updated 00DIST.

@lrosenman please also add more of the patches from the sysutils/lsof port. At least the namecache patch is needed on FreeBSD 12.2, otherwise lsof just hangs.

masatake:
I am not sure whether this works well or not.
However, merging may be better to get feedback from users.
Let's see what happens.

DamjanJovanovic:
> I am not sure whether this works well or not.
> However, merging may be better to get feedback from users.
> Let's see what happens.

Thank you :)

masatake:
I'm sorry to take a long time.

Issue #145, opened 2020-11-23
Compile and install the latest git master version of lsof on Ubuntu 20.04.


hongyi-zhao:
I try to compile and install the latest git master version of lsof on Ubuntu 20.04 with the folloiwng steps:

```
$ git clone https://github.com/lsof-org/lsof.git lsof-org/lsof.git
$ cd lsof-org/lsof.git/
$ sudo apt-get build-dep lsof
$ ./Configure linux
$ make
$ sudo make install
[...]

Please write your own install rule.  Lsof should be installed
setuid to root if you wish any lsof user to be able to examine
all open files.  Your install rule actions might look something
like this:

    install -m 4xxx -o root -g ${GRP} ${PROG} ${BIN}
    install -m 444 ${MAN} ${DOC}

You will have to complete the 4xxx modes, the GRP value, and
the skeletons for the BIN and DOC strings, given at the
beginning of this Makefile, e.g.,

    BIN= ${DESTDIR}/usr/local/etc
    DOC= ${DESTDIR}/usr/man/man8
    GRP= sys

$ lsof -v
lsof version information:
    revision: 4.93.2
    latest revision: https://github.com/lsof-org/lsof
    latest FAQ: https://github.com/lsof-org/lsof/blob/master/00FAQ
    latest (non-formatted) man page: https://github.com/lsof-org/lsof/blob/master/Lsof.8
    compiler: cc
    compiler version: 9.2.1 20200123 (Ubuntu 9.2.1-25ubuntu1) 
    compiler flags: -g -O2 -fdebug-prefix-map=/build/lsof-BXaWyU/lsof-4.93.2+dfsg=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -DLINUXV=54016 -DGLIBCV=230 -DHASIPv6 -DNEEDS_NETINET_TCPH -DHASSELINUX -DHASUXSOCKEPT -DHASPTYEPT -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -DHAS_STRFTIME -DLSOF_VSTR="5.4.16" -O
    loader flags: -L./lib -llsof -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now -lselinux
    Anyone can list all files.
    /dev warnings are disabled.
    Kernel ID check is disabled.
```
[...]

PR #153, opened 2021-04-22
Handle #145

Franklin-Qi:
For handle `Compile and install the latest git master version of lsof on Ubuntu 20.04. #145`

While contrasting to other manpages, such as iptables.8.gz, arp.8.gz and so on, i figure out `.so ./version` is not necessary.
Therefore,  delete `.so ./version` in Lsof.8 may be the best way to solve this problem, which i have tested successful in ubuntu 20.04.

### DIFF
--- a/Lsof.8
+++ b/Lsof.8
@@ -1,4 +1,3 @@
-.so ./version
 .TH LSOF 8 Revision-\*(VN
 \" Register )P is used neither by this file nor any groff macro.  However,
 \" some versions of nroff require it.


dilinger:
There's also ".so ./00DIALECTS" in the manpage on line 77. I don't know whether @masatake prefers removing the .so lines or running the manpage through soelim, but if we remove the '.so ./version' line it should probably also happen for '.so ./00DIALECTS' as well.

jiegec:
The correct way is to use `soelim`.

PR #158, opened 2021-06-14
Fix FD field description

algorythmic:
In https://github.com/lsof-org/lsof/commit/811dc78cc6404cb39d82bf01728cfdf38f097af1 the output format was changed to not select the `f` field by default, however the field description in `lsof_fields.h`, as seen in `-F?` output still included the `(always selected)` text.


masatake:
Thank you!

PR #161, opened 2021-06-20
Adjust alignment of buffer passed to stat()

### LINKED ISSUE #160: lsof crash on glibc 2.33 armv7a
Hi. Starting with glibc v2.33 the {f}stat{at} APIs got reworked which exposed a bug in lsof making it always crash on 32bit arm systems due to a buffer alignment problem in [this code location](https://github.com/lsof-org/lsof/blob/4.94.0/misc.c#L364).

The glibc 2.33 commit causing the crash this: aa03f722f3 linux: Add {f}stat{at} y2038 support

See the [discussion in the glibc bugtracker](https://sourceware.org/bugzilla/show_bug.cgi?id=27993) which uncovered the issue. Here's a stack trace:

```
#0  0xf396e1aa in __cp_stat64_t64_stat64 (st64_t64=0xffa4bc78, st64=<optimized out>)
    at ../sysdeps/unix/sysv/linux/stat_t64_cp.c:40
#1  0xf39671a8 in __GI___stat64 (file=<optimized out>, buf=0xffa4bd2b)
    at ../sysdeps/unix/sysv/linux/stat64.c:38
#2  0x066cbf26 in doinchild (fn=<optimized out>, fp=<optimized out>, rbuf=<optimized out>, 
    rbln=<optimized out>) at misc.c:364
#3  0x066cc09e in Readlink (arg=0x6b59378 "/proc") at misc.c:1109
#4  0x066bbe52 in readmnt () at dmnt.c:512
#5  0x066c6454 in ck_file_arg (i=1, ac=2, av=0xffa52224, fv=0, rs=0, sbp=0x0) at arg.c:189
#6  0x066ca506 in main (argc=<optimized out>, argv=<optimized out>) at main.c:1251
```

In a nutshell lsof needs to ensure the rbuf in doinchild() is properly aligned.

masatake:
Close #160.

The original code passes char[] buffer to stat().
This can be cause a SIGBUS.

#160 reported an actual crash on armv7a + glibc-2.33 platform.
See also https://sourceware.org/bugzilla/show_bug.cgi?id=27993.

The issue is reported by @10ne1.

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -554,8 +554,9 @@ provided test systems where I was able to do development work.
 	@albert-github
 	Tobias Geerinckx-Rice
 	Andres Salomon
-	Nicholas Bamber and
-	Grisha Levit
+	Nicholas Bamber
+	Grisha Levit and
+	@10ne1 (github account)
 
 If I have omitted a contributor's name, the fault is wholly mine,
 and I apologize for the error.
--- a/00DIST
+++ b/00DIST
@@ -5246,5 +5246,12 @@ July 14, 2018
 		Provided by @algorythmic (Grisha Levit) in #158.
 
 
+		Adjust alignment of buffer passed to stat().
+		The original code passes char[] buffer to stat(). This can be cause
+		a SIGBUS. #160 reported an actual crash on armv7a + glibc-2.33 platform.
+		See also https://sourceware.org/bugzilla/show_bug.cgi?id=27993.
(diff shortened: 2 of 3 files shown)


masatake:
circleci can be used for testing on arm.

masatake:
The original bug report is for Linux. However, the fix I proposed is not Linux-specific.  So I didn't put [linux] prefix.

PR #168, opened 2021-10-05
Manpage hyphen fixes

dilinger:
More Debian patch upstreaming, this time for the man page. As described in https://www.spinics.net/lists/linux-man/msg18858.html , command-line option hyphens in manpages should have the hyphen escaped so they render correctly in all character encodings. This ensures that they're all escaped properly, and also unescapes some hyphens that aren't used for command-line options. The tl;dr version of the above link is https://www.caichinger.com/blog/2006/09/06/man_hyphens/.


dilinger:
Woohoo, I remembered to actually edit 00DIST this time. 🎉

masatake:
Thank you very much.

PR #170, opened 2021-10-13
Upstream FreeBSD patches from the sysutils/lsof port

emaste:


masatake:
@lrosenman, @emaste, do you have more changes?
I'm thinking about making a new release.

@lrosenman, could you consider update 00DISTS and 00CREDITS?
00DISTS is needed to be updated if the changes are user visible.
00DISTS lists contributors.

PR #172, opened 2021-11-24
Configure: override cflags

ffontaine:
[Retrieved from: https://git.buildroot.net/buildroot/tree/package/lsof/0003-override-cflags.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


masatake:
Could you write more about your change?
It is not so obvious change for me.
Why do you need this?

masatake:
Thank you. Could you write it to the commit log?

My understanding is that this pull request fixes the broken LSOF_CFLAGS_OVERRIDE, a configure option.
Am I correct?

Could you update 00DIST and 00CREDITS?

ffontaine:
I updated the PR

PR #177, opened 2021-11-25
[FreeBSD] update 00DIST for configure change

lrosenman:


00DIST
```diff
@@ -5286,6 +5286,10 @@ July 14, 2018
                [FreeBSD] remove various old FreeBSD versions from support
                submitted by @emaste
 
+               [FreeBSD]onfigure: suggest variable to set if FreeBSD sys not
```
masatake, line 5289: onfigure? `c` may be missed.

lrosenman: fixed

PR #182, opened 2021-12-09
docs: clean up whitespaces in 00DIST

masatake:


masatake:
@lrosenman, could you use tab characters instead of space characters when adding entries to 00DIST?
Please, don't ask me why.

PR #194, opened 2022-03-10
lib/Makefile.skel: fix AR and RANLIB when cross-compiling

ffontaine:
This patch has been added in the context of buildroot sixteen years ago, here is an extract of
https://git.buildroot.net/buildroot/commit/package/lsof?id=aca398bb7fa956dc44e1442969c4a2a425780997:

> compile against the cross toolchain, not the host

[Retrieved from: https://git.buildroot.net/buildroot/tree/package/lsof/0001-makefile.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/00DIST
+++ b/00DIST
@@ -5293,6 +5293,9 @@ July 14, 2018
                Fix broken LSOF_CFLAGS_OVERRIDE.
                Provided by Fabrice Fontaine in #172.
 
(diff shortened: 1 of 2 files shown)


masatake:
Thank you.

PR #208, opened 2022-04-29
Various non-critical fixes

masatake:
Close #206.

I introduced -Wall option in this pull request.
As the result, we'll get more warnings.

PR #211, opened 2022-04-29
Clean up the call to fill_portmap

dilinger:
The pm variable is only used if !defined(HASNORPC_H), so we should
protect it with an #if. However, rather than doing that let's move
it to the fill_portmap function itself. One less #if to mentally
process.


masatake:
Merged via #208. Thank you.

PR #212, opened 2022-04-30
[linux] Avoid some easy collissions for udp/udp6 sockets when hashing

BenBE:
As promised in #208 here's that minor hash function change.

The values in the change are chosen in a way that similar values from different components that make up the hash are added with a slight shift relative to each other.

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -561,8 +561,9 @@ provided test systems where I was able to do development work.
 	@10ne1 (github account)
 	@a1346054
 	Damjan Jovanovic
-	Fabrice Fontaine and
-	Danny Fowler
+	Fabrice Fontaine
+	Danny Fowler and
+	Benny "BenBE" Baumann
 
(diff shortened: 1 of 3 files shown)


masatake (approved):

masatake:
@dilinger do you have a comment?

BenBE:
One comment I might add from my side is, that hashtables usually work best, when buckets are rarely hit. While debugging this I had a brief look into the raw data and saw, that `/proc/net/sockstat` reported >2200 used sockets on a moderately active system. With a hashmap size of 128 for the socket information this gives approx 20 sockets per bucket. Even when assuming an even distribution this causes the linear search aspect of the map implementation to be noticeable. Increasing the hashmap size to e.g. 4096 entries should mitigate this.

dilinger:
No comment, I'd have to really dig into it and don't have time right now. Sorry!

masatake:
Thank you!

PR #213, opened 2022-05-07
[linux] Changing the number of ipcbuckets to 4096

Alienmaster:
I suggest to increase the number of ipcbckets. On an fresh installed Ubuntu VM there are over 200 sockets used.
BenBE mentioned this problem here: https://github.com/lsof-org/lsof/pull/212#issuecomment-1113964661 .

I also fixed a small typo.


masatake:
Thank you. Could you add an entry for this change to the end of  `00DIST` file?
I think this is a user-visible change.

masatake:
Thank you.

PR #217, opened 2022-05-20
linux: Make build reproducible by default

danyspin97:
At openSUSE we have found out that lsof is still not completely reproducible due to the hostname being included in the resulting binary. This PR allows for reproducible hostname.

https://bugzilla.suse.com/show_bug.cgi?id=1199709

### DIFF
--- a/00DIST
+++ b/00DIST
@@ -5349,5 +5349,10 @@ July 14, 2018
 		[linux] Changing the number of ipcbuckets to 4096
 
 
+		[linux] Make build reproducible by checking SOURCE_DATE_EPOCH
+		and considering LSOF_{HOST,LOGNAME,SYSINFO,USER} as "none" when
+		it is set.
+		Provided by Danilo Spinella in #217
+
 The lsof-org team at GitHub
 April 28, 2022
--- a/dialects/linux/Makefile
+++ b/dialects/linux/Makefile
@@ -89,7 +89,11 @@ version.h:	FRC
 	@echo '#define	LSOF_CCFLAGS	"'`echo ${CFLAGS} | sed 's/\\\\(/\\(/g' | sed 's/\\\\)/\\)/g' | sed 's/"/\\\\"/g'`'"' >> version.h
 	@echo '#define	LSOF_CINFO	"${CINFO}"' >> version.h
 	@if [ "X${LSOF_HOST}" = "X" ]; then \
-	  echo '#define	LSOF_HOST	"'`uname -n`'"' >> version.h; \
+	  if [ "X${SOURCE_DATE_EPOCH}" = "X" ]; then \
+	    echo '#define	LSOF_HOST	"'`uname -n`'"' >> version.h; \
+	  else \
+	    echo '#define	LSOF_HOST	""' >> version.h; \
+	  fi \
 	else \
 	  if [ "${LSOF_HOST}" = "none" ]; then \
 	    echo '#define	LSOF_HOST	""' >> version.h; \
@@ -99,7 +103,11 @@ version.h:	FRC
 	fi
 	@echo '#define	LSOF_LDFLAGS	"${CFGL}"' >> version.h
 	@if [ "X${LSOF_LOGNAME}" = "X" ]; then \
-	  echo '#define	LSOF_LOGNAME	"${LOGNAME}"' >> version.h; \
+	  if [ "X${SOURCE_DATE_EPOCH}" = "X" ]; then \
+	    echo '#define	LSOF_LOGNAME	"${LOGNAME}"' >> version.h; \
+	  else \
+	    echo '#define	LSOF_LOGNAME	""' >> version.h; \
+	  fi \
 	else \
 	  if [ "${LSOF_LOGNAME}" = "none" ]; then \
 	    echo '#define	LSOF_LOGNAME	""' >> version.h; \
@@ -108,7 +116,11 @@ version.h:	FRC
 	  fi; \
 	fi
 	@if [ "X${LSOF_SYSINFO}" = "X" ]; then \
+	  if [ "X${SOURCE_DATE_EPOCH}" = "X" ]; then \
 	    echo '#define	LSOF_SYSINFO	"'`uname -a`'"' >> version.h; \
+	  else \
+	    echo '#define	LSOF_SYSINFO	""' >> version.h; \
+	  fi \
 	else \
 	  if [ "${LSOF_SYSINFO}" = "none" ]; then \
 	    echo '#define	LSOF_SYSINFO	""' >> version.h; \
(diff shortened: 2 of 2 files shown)


bmwiedemann:
need to patch `uname` calls in `dialects/linux/Makefile`

bmwiedemann:
The `uname -a` 20 lines below is still there and includes the hostname as well.

danyspin97:
Yes, but I can't think of any sane value to replace it with. I can just set `LSOF_SYSINFO=none` when running make instead.

bmwiedemann:
You could also replace it with `uname -rv` - either always (for simplicity) or if `$SOURCE_DATE_EPOCH` is set.

masatake:
I think "reproducible build" issue was fixed in #150.

Isn't `make LSOF_HOST=none ` enough?

bmwiedemann:
`make LSOF_SYSINFO=none LSOF_HOST=none`
makes it reproducible.

Would be nice to either document that better than `00README`
```
Note in the General Environment Variables section of the 00XCONFIG
file that there are five environment variables that can be used to
pre-define values in lsof's -v output: LSOF_BLDCMT, LSOF_HOST,
LSOF_LOGNAME, LSOF_SYSINFO, and LSOF_USER.
```

or to have a way for it to become reproducible by default when `$SOURCE_DATE_EPOCH` indicates that reproducibility is important.

danyspin97:
I agree that being reproducible by default is important when `SOURCE_DATE_EPOCH` is set.

masatake:
> make LSOF_SYSINFO=none LSOF_HOST=none

Isn't this command line enough?

danyspin97:
> Isn't this command line enough?

It is but I don't like it. Why should those info be disabled for reproducible builds (which are used by many distributions by default)? I could argue that it either makes sense to have them, or makes sense to remove them. I would also argue that enabling reproducible builds by settings some obscure feature manually is not ideal either.

masatake:
> It is but I don't like it. Why should those info be disabled for reproducible builds (which are used by many distributions by default)?

`none` is for disabling. However, you can specify any strings:
```
$ make LSOF_SYSINFO=ABC LSOF_HOST=builder.example.com
...
$ ./lsof -v       
lsof version information:
    revision: 4.95.0
    latest revision: https://github.com/lsof-org/lsof
    latest FAQ: https://github.com/lsof-org/lsof/blob/master/00FAQ
    latest (non-formatted) man page: https://github.com/lsof-org/lsof/blob/master/Lsof.8
    constructed by and on: yamato@builder.example.com
    compiler: cc
    compiler version: 11.3.1 20220421 (Red Hat 11.3.1-2) (GCC) 
    compiler flags: -DLINUXV=517006 -DGLIBCV=234 -DHASIPv6 -DHASNORPC_H -DNEEDS_NETINET_TCPH -DHASSELINUX -DHASUXSOCKEPT -DHASPTYEPT -DHASSOSTATE -DHASSOOPT -D_FILE_OFFSET_BITS=64 -D_LARGEFILE64_SOURCE -DHAS_STRFTIME -DLSOF_VSTR="5.17.6" -O
    loader flags: -L./lib -llsof  -lselinux
    system info: ABC
```

Passing specific values via parameters defined in Makefile is an ordinary way.


> I don't like it.

Is this the reason you made this pull request?

The issue you want to solve is unclear to me.
So I cannot preview the change.

>  Allow setting hostname for reproducible builds #217 

This is the title of this pull request.
As I showed, hostname can be set without your change.
So I cannot find the reason to merge this.

danyspin97:
> Is this the reason you made this pull request?

I made this PR because I wanted to make reproducible builds by default. I have updated the title to make this more clear.

> Passing specific values via parameters defined in Makefile is an ordinary way.

Except they are also set in the Configure and you would think they are not set again in the Makefile. This is not ordinary by any means.

>  So I cannot find the reason to merge this.

Passing those arguments is a workaround, they should have been read and set accordingly in the first place and this PR just fixes that. I understand you can "do" that manually, but this saves time for anyone else using reproducible builds (e.g. debian just to name one); it makes the openSUSE package easier and helps anyone contributing to it in the future.

dilinger:
So as I'm understanding this, the goal is to avoid doing the kind of thing that we currently do in Debian, which is this:
```
# prevent recording of details of the build environment to ensure build
# reproducibility
LSOF_CCDATE := none
LSOF_HOST := none
LSOF_LOGNAME := none
LSOF_SYSINFO := none
LSOF_USER := none
export LSOF_CCDATE LSOF_HOST LSOF_LOGNAME LSOF_SYSINFO LSOF_USER
```

The LSOF_SYSINFO proposed change uses SOURCE_DATE_EPOCH (documented at https://reproducible-builds.org/docs/source-date-epoch/ ) as a global flag to determine that a reproducible build is desired, and to override LSOF_SYSINFO with whatever's in SOURCE_DATE_EPOCH.  I don't mind using SOURCE_DATE_EPOCH like that, but I don't particularly like setting LSOF_SYSINFO to the value of SOURCE_DATE_EPOCH; I'd rather see the code check if LSOF_SYSINFO = "none" or SOURCE_DATE_EPOCH is set, then use the existing ```echo '#define	LSOF_SYSINFO	""' >> version.h;```.  Otherwise, when a distribution user runs ```lsof -v```, they're going to see "system info: " and then some random string from SOURCE_DATE_EPOCH.

I'd actually suggest taking this a step further, and (somewhere at the beginning of Configure) checking if SOURCE_DATE_EPOCH is set, and if so overriding all of the above variables (LSOF_{HOST,LOGNAME,SYSINFO,USER}) to "none". That removes the need to special-case LSOF_SYSINFO versus the other unreproducable variables. It would result in the inability to set both SOURCE_DATE_EPOCH and, say, LSOF_HOST=special.host, but that seems like a weird corner case that I doubt anyone cares about.

masatake:
As I assigned @dilinger to this pull request, I don't have a strong opinion about this topic. If you can reduce 'if' branches, it is nice.
If you will add an 'if' branch, we need a rather strong reason. Updating documents are welcome.

danyspin97:
> I'd actually suggest taking this a step further, and (somewhere at the beginning of Configure) checking if SOURCE_DATE_EPOCH is set, and if so overriding all of the above variables (LSOF_{HOST,LOGNAME,SYSINFO,USER}) to "none". That removes the need to special-case LSOF_SYSINFO versus the other unreproducable variables. It would result in the inability to set both SOURCE_DATE_EPOCH and, say, LSOF_HOST=special.host, but that seems like a weird corner case that I doubt anyone cares about.

This doesn't work, because variables set in the Configure are not actually read by the Makefile.

danyspin97:
> I'd rather see the code check if LSOF_SYSINFO = "none" or SOURCE_DATE_EPOCH is set, then use the existing echo '#define LSOF_SYSINFO ""' >> version.h;

I have applied this to LSOF_{HOST,LOGNAME,SYSINFO,USER} and now the build should be fully reproducible by default without exporting anything else other than `SOURCE_DATE_EPOCH` (so that the lines taken from the debian package could be entirely removed).

dilinger:
Nice, I like that better. I just noticed the change for LSOF_SYSINFO from 'uname -a' to 'uname -rv'; I'm assuming that was from before to make builds reproducible without including a hostname. It's probably fine to leave it as 'uname -a' now though, since it won't be used at all when SOURCE_DATE_EPOCH is set.

Also, you can go ahead and add your name to the changelog entry in 00DIST (only masatake doesn't add his name).

dilinger:
> > Nice, I like that better. I just noticed the change for LSOF_SYSINFO from 'uname -a' to 'uname -rv'; I'm assuming that was from before to make builds reproducible without including a hostname. It's probably fine to leave it as 'uname -a' now though, since it won't be used at all when SOURCE_DATE_EPOCH is set.
> 
> You're right, this change isn't needed anymore, reverted.
> 
> > Also, you can go ahead and add your name to the changelog entry in 00DIST (only masatake doesn't add his name).
> 
> Done this too 👍🏻

Thanks! @masatake , this looks good to me.

masatake:
Thank you.

PR #221, opened 2022-06-10
darwin: fix build with -fno-common (Cfp redefinition)

trofi:
gcc-10 and llvm-11 changed the default from -fcommon to -fno-common:
  https://gcc.gnu.org/PR85678

As a result build fails as:

    duplicate symbol '_Cfp' in: ddev.o dfile.o

`Cfp` is already explicitly defined in `dstore.c`. The change turns
header definitino into declaration.


masatake:
@dannykansas, I will merge this.

masatake:
@trofi, thank you.

PR #226, opened 2022-07-13
darwin: fix global variable definition in dlsof.h

jiegec:
Defining global variables in headers is error prone and can break
builds with -fno-common flag enabled:

    duplicate symbol '_Cfp' in:
    ddev.o
    dfile.o

The issue is fixed in libproc backend in the same way as pr #221.


masatake (approved):

masatake:
Thank you. Perfect. I used `[darwin]` instead of `drawin` as the prefix. However, it is a minor thing.

If your change (or a contributor's change you will review in the future) is a user-visible one, please, update 00DIST, too.

Please, watch what I will do after accepting a contribution...

Issue #238, opened 2022-07-20
[BUG] list memory mapped files returns with 'stat: No such file or directory'


Hunter1016:
**Describe the bug**
list memory mapped files in different mount namespace return with 'stat: No such file or directory'.

**To Reproduce**
1. Run 'lsof -p <pid>', this pid in a different mount namespace from lsof itself, for example, a container process.
2. See error

**Expected behavior**
Should not print 'stat: No such file or directory'.

**Program output**
COMMAND     PID USER   FD   TYPE DEVICE SIZE/OFF    NODE NAME
bash    2307034 root  cwd    DIR  0,104     4096 2233949 /root
bash    2307034 root  rtd    DIR  0,104     4096 2361890 /
bash    2307034 root  txt    REG  0,104   964536 2234000 /usr/bin/bash
bash    2307034 root  mem    REG  254,1          2234000 /usr/bin/bash (path inode=263609)
bash    2307034 root  mem    REG  254,1          2236028 /usr/lib64/libnss_files-2.17.so (stat: No such file or directory)
bash    2307034 root  mem    REG  254,1          2235885 /usr/lib64/libc-2.17.so (stat: No such file or directory)
bash    2307034 root  mem    REG  254,1          2235914 /usr/lib64/libdl-2.17.so (stat: No such file or directory)
bash    2307034 root  mem    REG  254,1          2236135 /usr/lib64/libtinfo.so.5.9 (stat: No such file or directory)
bash    2307034 root  mem    REG  254,1          2235861 /usr/lib64/ld-2.17.so (stat: No such file or directory)
bash    2307034 root    0u   CHR  136,0      0t0       3 /dev/pts/0
bash    2307034 root    1u   CHR  136,0      0t0       3 /dev/pts/0
bash    2307034 root    2u   CHR  136,0      0t0       3 /dev/pts/0
bash    2307034 root  255u   CHR  136,0      0t0       3 /dev/pts/0

**Environment (please complete the following information):**

 - Kernel: 4.19
 - OS: Debian
 - lsof Version: 4.91

Hunter1016:
#239

Issue #126, opened 2020-10-29
[linux] handling mount namespace


masatake:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=731780

PR #239, opened 2022-07-20
[linux] obtain correct information of memory-mapped file.

Hunter1016:
In circumstances the target process in different mount namespace
from lsof itself, applying stat(2) directly gets failed,
with error message 'stat: No such file or directory'.

Thus, fork a child process and move it into the mount namespace
of the target process before applying stat(2). In this way, correct
information can be obtained.

Signed-off-by: huteng.ht <huteng.ht@bytedance.com>
#238

### DIFF
--- a/dialects/linux/dproc.c
+++ b/dialects/linux/dproc.c
@@ -69,6 +69,9 @@
 #define	ULLONG_MAX		18446744073709551615ULL
 #endif	/* !defined(ULLONG_MAX) */
 
+#define	NS_PATH_LENGTH		100     /* namespace path string length */
+#define	MAP_PATH_LENGTH		100     /* map_files path length */
+#define	ADDR_LENGTH		100     /* addr range of map_files length */
 
 /*
  * Local structures
@@ -1395,6 +1398,31 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
 	return(0);
 }
 
+/* compare mount namespace of this lsof process and the target process */
+
+static int
+compare_mntns(pid)
+        int pid;                        /* pid of the target process */
+{
+	char nspath[NS_PATH_LENGTH];
+	struct stat sb_self, sb_target;
+	int ret;
+
+	if (stat("/proc/self/ns/mnt", &sb_self))
+	    return -1;
+
+	ret = snprintf(nspath, sizeof(nspath), "/proc/%d/ns/mnt", pid);
+	if (ret >= sizeof(nspath) || ret <= 0)
+	    return -1;
+
+	if (stat(nspath, &sb_target))
+	    return -1;
+
+	if (sb_self.st_ino != sb_target.st_ino)
+	    return -1;
+
+	return 0;
+}
 
 /*
  * process_proc_map() - process the memory map of a process
@@ -1425,12 +1453,18 @@ process_proc_map(p, s, ss)
 	static int sma = 0;
 	static char *vbuf = (char *)NULL;
 	static size_t vsz = (size_t)0;
+	int diff_mntns = 0;
 /*
  * Open the /proc/<pid>/maps file, assign a page size buffer to its stream,
  * and read it/
  */
 	if (!(ms = open_proc_stream(p, "r", &vbuf, &vsz, 0)))
 	    return;
+
+	/* target process in a different mount namespace from lsof process. */
+	if (compare_mntns(Lp->pid))
+	    diff_mntns = 1;
+
 	while (fgets(buf, sizeof(buf), ms)) {
 	    if ((nf = get_fields(buf, ":", &fp, &eb, 1)) < 7)
 		continue;			/* not enough fields */
@@ -1518,11 +1552,44 @@ process_proc_map(p, s, ss)
 		efs = sv = 1;
 	    else
 		efs = 0;
-	    if (!efs) {
-		if (HasNFS)
-		    sv = statsafely(fp[6], &sb);
-		else
-		    sv = stat(fp[6], &sb);
+
+	    /* For processes in different mount namespace from lsof process,
+	     * stat corresponding files under /proc/[pid]/map_files would follow
+	     * symlinks regardless of namespaces.
+	     */
+	    if (diff_mntns) {
+		char path[MAP_PATH_LENGTH];
+		char addr[ADDR_LENGTH];
+		uint64_t start, end;
+		int ret;
+
+		if (sscanf(fp[0], "%lx-%lx", &start, &end) != 2)
+		    goto stat_directly;
+
+		ret = snprintf(addr, sizeof(addr), "%lx-%lx", start, end);
+		if (ret >= sizeof(addr) || ret <= 0)
+		    goto stat_directly;
+
+		ret = snprintf(path, sizeof(path), "/proc/%d/map_files/%s",
+			       Lp->pid, addr);
+		if (ret >= sizeof(path) || ret <= 0)
+		    goto stat_directly;
+
+		if (!efs) {
+		    if (HasNFS)
+		        sv = statsafely(path, &sb);
+		    else
+		        sv = stat(path, &sb);
+	        }
+	    }
+	    else {
+stat_directly:
+		if (!efs) {
+		    if (HasNFS)
+			sv = statsafely(fp[6], &sb);
+		    else
+			sv = stat(fp[6], &sb);
+		}
 	    }
 	    if (sv || efs) {
 		en = errno;
@@ -1547,6 +1614,8 @@ process_proc_map(p, s, ss)
 		    nmabuf[sizeof(nmabuf) - 1] = '\0';
 		    (void) add_nma(nmabuf, strlen(nmabuf));
 		}
+	    } else if (diff_mntns) {
+		mss = SB_ALL;
 	    } else if ((sb.st_dev != dev) || ((INODETYPE)sb.st_ino != inode)) {
 
 	    /*
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       - run:
           name: Install build tools
           command: |
-            dnf -y install git gcc make sudo nmap-ncat iproute libtirpc-devel lcov gem
+            dnf -y install git gcc make sudo nmap-ncat iproute libtirpc-devel lcov gem procps
             gem install lcoveralls
       - checkout
       - run:
@@ -40,7 +40,7 @@ jobs:
       - run:
           name: Install build tools
           command: |
-            dnf -y install git gcc make sudo nmap-ncat iproute libtirpc-devel
+            dnf -y install git gcc make sudo nmap-ncat iproute libtirpc-devel procps
       - checkout
       - run:
           name: Configure
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,4 @@ dialects/linux/tests/pidfd
 dialects/linux/tests/pipe
 dialects/linux/tests/pty
 dialects/linux/tests/ux
+dialects/linux/tests/mmap
--- a/00DIST
+++ b/00DIST
@@ -5372,6 +5372,8 @@ July 14, 2018
 		Fixes issue #234.
 		Provided by Jiajie Chen in #235
 
+		[linux] obtain correct information of memory-mapped file.
+		Provided by Teng Hu in #239
 
 The lsof-org team at GitHub
 April 28, 2022
--- a/dialects/linux/tests/Makefile
+++ b/dialects/linux/tests/Makefile
@@ -7,6 +7,7 @@ HELPERS = \
 	pipe \
 	pty \
 	ux \
+	mmap \
 	\
 	open_with_flags \
 	\
--- a/dialects/linux/tests/case-20-mmap.bash
+++ b/dialects/linux/tests/case-20-mmap.bash
@@ -1,0 +1,61 @@
+#!/bin/sh
+
+lsof=$1
+report=$2
+tdir=$3
+
+cleanup()
+{
+    if [ -n "$(mount | grep "tmp\/ext4.img" | grep -v grep)" ]
+    then
+        umount /tmp/ext4.img
+    fi
+    
+    if [ -f /tmp/ext4.img ]
+    then
+        rm -f /tmp/ext4.img
+    fi
+    
+    if [ -d /tmp/TEST ]
+    then
+        rm -rf /tmp/TEST
+    fi
+    
+    pidlist=`ps aux | grep "TEST\/MMAP" | grep -v grep | awk '{print $2}'`
+    for pid in $pidlist
+    do
+        kill -9 $pid
+    done
+}
+
+# cleanup the environment
+cleanup
+
+# create a new namespace and mmap
+unshare --mount --propagation private $3/mount-and-mmap.bash $@ >> $report 2>&1
+if grep -q 'unshare failed: Operation not permitted' $report; then
+    echo "unshare is not supported on this platform" >> $report
+    exit 2
+fi
+
+# get pid of the mmap process
+pid=`ps aux | grep "TEST\/MMAP" | grep -v grep | awk '{print $2}'`
+if [ -z "$pid" ]
+then
+    echo "mmap process does not exist" >> $report
+    cleanup
+    exit 1
+fi
+
+# lsof -p pid and obtain the output
+output=`$lsof -p $pid | grep "TEST\/MMAP" | grep "stat: No such file or directory"`
+
+if [ -z "$output" ]
+then
+    cleanup
+    exit 0
+else
+    echo "unexpected output: $output" >> $report
+    cleanup
+    exit 1
+fi
--- a/dialects/linux/tests/mmap.c
+++ b/dialects/linux/tests/mmap.c
@@ -1,0 +1,33 @@
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#define handle_error(msg) \
+	do { perror(msg); exit(EXIT_FAILURE); } while (0)
+
+int main(int argc, char *argv[])
+{
+	char *addr;
+	int fd;
+	struct stat sb;
+
+	if (argc != 2)
+		exit(EXIT_FAILURE);
+
+	fd = open(argv[1], O_RDONLY);
+	if (fd == -1)
+		handle_error("open");
+
+	if (fstat(fd, &sb) == -1)
+		handle_error("fstat");
+
+	addr = mmap(NULL, sb.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+	if (addr == MAP_FAILED)
+		handle_error("mmap");
+
+	pause();
+	return 0;
+}
--- a/dialects/linux/tests/mount-and-mmap.bash
+++ b/dialects/linux/tests/mount-and-mmap.bash
@@ -1,0 +1,20 @@
+#!/bin/sh
+set -e
+
+lsof=$1
+report=$2
+tdir=$3
+
+TARGET=$tdir/mmap
+if ! [ -x $TARGET ]; then
+    echo "target executable ( $TARGET ) is not found" >> $report
+    exit 1
+fi
+
+mkdir /tmp/TEST
+dd if=/dev/zero of=/tmp/ext4.img bs=40MB count=1
+mkfs.ext4 /tmp/ext4.img
+mount --make-private /tmp/ext4.img /tmp/TEST
+
+dd if=/dev/zero of=/tmp/TEST/MMAP bs=2MB count=1
+$TARGET /tmp/TEST/MMAP &


dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,71 @@ process_proc_map(p, s, ss)
...
 	    else
 		efs = 0;
-	    if (!efs) {
-		if (HasNFS)
-		    sv = statsafely(fp[6], &sb);
-		else
-		    sv = stat(fp[6], &sb);
-	    }
+
+	/*
```
jiegec, line 1526: Align comments to code

dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,71 @@ process_proc_map(p, s, ss)
...
+	 * Fork a child process and create a pipe.
+	 */
+            if (pipe(pipefd) < 0) {
+                fprintf(stderr, "%s: can't open pipes: %s\n",
+                    Pn, strerror(errno));
+                Exit(1);
```
jiegec, line 1532: Is this error fatal? Can we fallback to our original code when pipe() fails?

dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,72 @@ process_proc_map(p, s, ss)
...
+
+                close(pipefd[0]);
+                sprintf(nspath, "/proc/%d/ns/mnt", Lp->pid);
+                nsfd = open(nspath, O_RDONLY);
+
+                if (nsfd < 0) {
+                    fprintf(stderr, "%s: can't open ns: %s\n",
+                        Pn, strerror(errno));
+                }
+                setns(nsfd, CLONE_NEWNS);
```
jiegec, line 1559: This is available since Linux 3.8, should we skip this feature on earlier kernels?

dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,72 @@ process_proc_map(p, s, ss)
...
+	     * Read the return code and stat buffer from pipefd[0].
+	     */
+                close(pipefd[1]);
+                if (read(pipefd[0], &sv, sizeof(sv)) != sizeof(sv) ||
+                    read(pipefd[0], &sb, sizeof(sb)) != sizeof(sb)) {
+                    fprintf(stderr, "%s: can't read from pipe: %s\n",
+                        Pn, strerror(errno));
+                    Exit(1);
+		}
+                close(pipefd[0]);
```
jiegec, line 1588: Do we need to wait for the client process to exit?

dialects/linux/dproc.c
Hunter1016: Helpful suggestion, lack of consideration, I'll amend it.

dialects/linux/dproc.c
Hunter1016: Yes, thanks, let me amend it.

dialects/linux/dproc.c
jiegec: I am unsure about the code style, @masatake

masatake: The style is not well defined. However, if the lines you changed are not well aligned with the other code around you changed ON GITHUB WEB INTERFACE, I would you to adjust your changes. In the other words, keep your change as if the change has been there for years.

masatake:
I would like to add some words.
This is related to https://github.com/lsof-org/lsof/issues/126 .

Handling namespaces in lsof was one of I wanted to work on.
However, I mostly stepped down from lsof development. Instead, I started developing lsfd (https://manpages.debian.org/testing/util-linux-extra/lsfd.1.en.html).

As far as I can remember correctly, lsfd uses /proc/$pid/map_files as the fallback for entries in /proc/$pid/maps.

Hunter1016:
Based on review opinions above, I've forced-updated this patch, please kindly review. @masatake @jiegec

misc.c
```diff
@@ -1699,3 +1700,40 @@ x2dev(s, d)
...
+}
+
+int
+compare_kernel_release(release)
+	char *release;          /* kernel release */
+{
+	struct utsname uts;
+
+	uname(&uts);
+	return strcmp(uts.release, release);
```
jiegec, line 1738: strcmp() may not give correct results for versions, e.g.: `3.8.0` > `3.11.0`

misc.c
```diff
@@ -1699,3 +1700,40 @@ x2dev(s, d)
 	*d = r;
 	return(s);
 }
+
+int
+compare_namespace(ns)
+	char *ns;                /* namespace */
+{
+	char nspath_self[100], nspath_target[100];
```
jiegec, line 1708: Do not write magic number `100`. Use definitions.

dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,90 @@ process_proc_map(p, s, ss)
...
+                        fprintf(stderr, "%s: can't setns: %s\n",
+                            Pn, strerror(errno));
+			Exit(LSOF_ERROR);
+                    }
+
+                    if (!efs) {
+                        if (HasNFS)
+                            sv = statsafely(fp[6], &sb);
+                        else
+                            sv = stat(fp[6], &sb);
```
jiegec, line 1566: Should we add some flag to tell the user that the file actually lies in another mount namespace? It can be confusing for the user to see that a process opens a file but the user cannot find that file in the default namespace.

dialects/linux/dlsof.h
```diff
@@ -40,7 +40,6 @@
 
 #include <dirent.h>
 #define	DIRTYPE	dirent			/* for arg.c's enter_dir() */
-#define	__USE_GNU			/* to get all O_* symbols in fcntl.h */
```
jiegec, line 43: This is required for `case-20-open-flags-path` to work.

dialects/linux/dproc.c
```diff
@@ -1518,12 +1522,90 @@ process_proc_map(p, s, ss)
...
+		    /* Wait the child process exit */
+		    waitpid(pid, &status, 0);
+
+		    /* If child process exit non-zero, goto stat_directly */
+		    if (WEXITSTATUS(status) != 0)
+			goto stat_directly;
+                }
+	    } else {
+stat_directly:
+                if (HasNFS)
```
jiegec, line 1603: `if (!efs)` is missing here.

Hunter1016: Thanks a lot.

dialects/linux/dlsof.h
Hunter1016: Thanks, I'll add a #ifndef __USE_GNU before this.

misc.c
Hunter1016: Would use LINUX_VERSION_CODE instead.

misc.c
Hunter1016: Got it.

dialects/linux/dproc.c
Hunter1016: Got it, thanks.

dialects/linux/dproc.c
Hunter1016: I considered add a flag, e.g. 'in another mnt namespace' and append it to nmabuf. Since current code on nmabuf is a little redundant, I prefer to do it in another patch focused on nmabuf handling.

Hunter1016:
@jiegec forced-updated, please kindly review. The runner fedora36 'Report coverage' stage has failed, but I cannot get why from the error log.

jiegec (approved):

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,50 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+
+int
+compare_namespace(ns)
+        char *ns;                /* namespace */
+{
+        char nspath_self[NS_PATH_LENGTH], nspath_target[NS_PATH_LENGTH];
+        char ns_self[NS_LENGTH], ns_target[NS_LENGTH];
+        ssize_t n_self, n_target;
+
+        sprintf(nspath_self, "/proc/self/ns/%s", ns);
```
BenBE, line 1410: May overflow the buffer if `ns` is sufficiently long. Prefer `snprintf` and check return value.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,50 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+        n_self = readlink(nspath_self, ns_self, NS_LENGTH);
+        if (n_self < 0)
+            return -1;
+
+        sprintf(nspath_target, "/proc/%d/ns/%s", Lp->pid, ns);
```
BenBE, line 1415: Similar buffer overflow issue here.

dialects/linux/dproc.c
Hunter1016: Since '/proc/self/ns/[ns]' is unlikely to be longer than NS_PATH_LENGTH, which is defined as 100, I'm afraid there's no need to.

dialects/linux/dproc.c
masatake: I also thought sprintf with no limit could be an issue.
However, I could not find a threat situation.

I have just found one:

```
$ mkdir x
$ mkdir x/lib64
$ sudo mount -o bind /lib64 x/lib64
$ mkdir x/lib                      
$ sudo mount -o bind /lib x/lib    
$ mkdir x/bin                  
$ sudo mount -o bind /bin x/bin
$ sudo chroot x /bin/bash
bash-5.1# mkdir /proc
bash-5.1# mkdir /proc/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/
bash-5.1# mkdir /proc/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/ns
bash-5.1# touch /proc/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/ns/mnt
bash-5.1# ls -l /proc/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/ns/mnt
-rw-r--r--. 1 0 0 0 Jul 23 04:40 /proc/0123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789/ns/mnt
```

With overlayfs, we can make a more realistic example.
Using sprintf is better.

Hunter1016: Considered on this situation, it's necessary and practical to handle this with snprintf. Thanks a lot.

Hunter1016:
forced-updated, possible overflow discussed above has been handled , please kindly review @masatake .

dialects/linux/dproc.c
BenBE: Depending on the threat model you should …

There's nothing stopping an attacker from mounting a malicious procfs over normal proc and generating excessively large names inside that directory …

Also it's just bad style leaving your assumptions unchecked when doing so can result in bugs like buffer overflows (which in this case could potentially allow for arbitrary code execution).

dialects/linux/dproc.c
```diff
@@ -28,7 +28,8 @@
  *
  * 4. This notice may not be removed or altered.
  */
-
```
masatake, line 31: Removing an empty line is nothing to do with extracting namespace information.

dialects/linux/dproc.c
```diff
@@ -69,7 +70,8 @@
 #define	ULLONG_MAX		18446744073709551615ULL
 #endif	/* !defined(ULLONG_MAX) */
 
-
```
masatake, line 72: Removing an empty line is nothing to do with extracting namespace information.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+int
+compare_namespace(ns)
+        char *ns;                /* namespace */
+{
+        char nspath_self[NS_PATH_LENGTH], nspath_target[NS_PATH_LENGTH];
+        char ns_self[NS_LENGTH], ns_target[NS_LENGTH];
+        ssize_t n_self, n_target;
+        int ret;
+
+        ret = snprintf(nspath_self, NS_PATH_LENGTH, "/proc/self/ns/%s", ns);
```
masatake, line 1411: Instead of using `NS_PATH_LENGTH`, you can use `sizeof nspath_self`.
I wonder which one is better.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+        if (ret >= NS_PATH_LENGTH || ret <= 0)
+	    return -1;
+
+        n_self = readlink(nspath_self, ns_self, NS_LENGTH);
+        if (n_self < 0)
+            return -1;
+
+        ret = snprintf(nspath_target, NS_PATH_LENGTH, "/proc/%d/ns/%s", Lp->pid, ns);
+        if (ret >= NS_PATH_LENGTH || ret <= 0)
+	    return -1;
```
masatake, line 1421: As far as reading this line on a web browser, `if` and `line` are alinged.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+        char *ns;                /* namespace */
+{
+        char nspath_self[NS_PATH_LENGTH], nspath_target[NS_PATH_LENGTH];
+        char ns_self[NS_LENGTH], ns_target[NS_LENGTH];
+        ssize_t n_self, n_target;
+        int ret;
+
+        ret = snprintf(nspath_self, NS_PATH_LENGTH, "/proc/self/ns/%s", ns);
+        if (ret >= NS_PATH_LENGTH || ret <= 0)
+	    return -1;
```
masatake, line 1413: As far as reading this line on a web browser, `if` and `line` are alinged.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+
+        n_self = readlink(nspath_self, ns_self, NS_LENGTH);
```
masatake, line 1415: Instead of using `NS_LENGTH`, you can use `sizeof ns_self`.
 I wonder which one is better.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
 	return(0);
 }
 
+/* compare namespace of this lsof proces and the target process */
+
+int
+compare_namespace(ns)
```
masatake, line 1403: How about passing `pid`, instead of referring `Lp->pid`?

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+        if (strncmp(ns_self, ns_target, n_self))
+            return -1;
+
+        return 0;
+}
+
+/* compare kernel release */
+
+int
+compare_kernel_release(release)
```
masatake, line 1439: Do you want to do this in runtime or build time?

A funcitn is for runtime.
A macro is for build-time.

`LINUXV` is defined in build-time.
`release` is always 38000. So we can evaluate a result at build-time.
Defining this with a macro, you can improve the performance a bit.

However, what we would like to know is a fact at runtime.
Consider, a linux distribution ships a lsof binary package.
The distributor has a build farm that uses linux-3.7.
The new version of the distribution includes linux-5.x.
The runtime where a lsof binary runs is linux-5.x.
However, the lsof is built on linux-3.7.

dialects/linux/dproc.c
```diff
@@ -1425,6 +1478,9 @@ process_proc_map(p, s, ss)
 	static int sma = 0;
 	static char *vbuf = (char *)NULL;
 	static size_t vsz = (size_t)0;
+        int pipefd[2];
```
masatake, line 1481: Unaligned as far as seen on my web browser.

dialects/linux/dproc.c
```diff
@@ -1518,12 +1574,91 @@ process_proc_map(p, s, ss)
 		efs = sv = 1;
 	    else
 		efs = 0;
-	    if (!efs) {
-		if (HasNFS)
-		    sv = statsafely(fp[6], &sb);
-		else
-		    sv = stat(fp[6], &sb);
+
```
masatake, line 1577: I don't understand why you added your code that introduces extra overhead here.

Isn't it enough to run your code only when `stat` is failed?

dialects/linux/dproc.c
```diff
@@ -1518,12 +1574,91 @@ process_proc_map(p, s, ss)
...
+	    /* kernel version >= 3.8.0 and the target process has a different
+	     * mount namespace with this lsof process. */
+	    if (compare_kernel_release(38000) && compare_namespace("mnt")) {
```
masatake, line 1580: Do we have to do this every time when extracting information from a mmap entry?
I think we should do this ONLY once with RUNTIME information.
uname(2) is for getting runtime information.

dialects/linux/dproc.c
```diff
@@ -1518,12 +1574,91 @@ process_proc_map(p, s, ss)
...
+                    char nspath[NS_PATH_LENGTH];
+
+                    close(pipefd[0]);
+                    sprintf(nspath, "/proc/%d/ns/mnt", Lp->pid);
+                    nsfd = open(nspath, O_RDONLY);
+
+                    if (nsfd < 0 || setns(nsfd, CLONE_NEWNS)) {
+                        fprintf(stderr, "%s: can't setns: %s\n",
+                            Pn, strerror(errno));
+			Exit(LSOF_ERROR);
```
masatake, line 1610: Unaligned on my web browser.

masatake: Do we have to report this?
The original 'stat: No such file or directory' looks enough.

dialects/linux/dproc.c
masatake: I understand the situation where this is not so simple.

I think the following algorithm is better.

1. get a line from /proc/$pid/maps. Let's call it $line.
2. get a file name from the $line. Let's call it $fname.
3. Call stat($fname). Let's call the result of stat buf $stat. If the syscall returns an error, let's use your new code.
4. get a (major,minor,inode) tuple from $line. Let's call the tuple $INODE.
5. compare $INODE with $stat. If they don't match, let's use your new code.
6. we confirmed $stat includes enough useful information. We don't have to run your new code.

Using fork & pipe takes costs. If just using stat is enough, we don't have to take the costs.

masatake:
I made some comments. In addition, we need a test case and an entry for 00DISTS.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
 	return(0);
 }
 
+/* compare namespace of this lsof proces and the target process */
+
+int
```
masatake, line 1402: `static` modifier can be added.

dialects/linux/dproc.c
```diff
@@ -1395,6 +1397,57 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
...
+
+        if (strncmp(ns_self, ns_target, n_self))
+            return -1;
+
+        return 0;
+}
+
+/* compare kernel release */
+
+int
```
masatake, line 1438: `static` modifier can be added.

dialects/linux/dproc.c
Hunter1016: Helpful, passing pid is more flexible.

dialects/linux/dproc.c
Hunter1016: Helpful, sizeof is better because it's more readable and alterable.

dialects/linux/dproc.c
Hunter1016: Helpful, sizeof is better because it's more readable and alterable.

dialects/linux/dproc.c
Hunter1016: Yes, it's my carelessness to mix build-time and runtime. This is a typical runtime case, use uname(2) instead.

dialects/linux/dproc.c
Hunter1016: > Do we have to report this? The original 'stat: No such file or directory' looks enough.

I think print error message can help to analyze why 'stat: No such file or directory' happens.

dialects/linux/dproc.c
Hunter1016: Will do this before the while loop.

dialects/linux/dproc.c
Hunter1016: This algorithm has less intrusion to current code, but I think logically, it seems not better.
According to this steps, we should always call stat(2) first of all, when we get failed, turn to the new code to try again. But **we can know it will surely fail** by namespace comparison in advance, so we don't need to call stat(2) firstly, and thus, cost of this syscall is saved.
About fork and pipe costs, this patch mainly obtain correct stat result at the expense of these costs for the processes in different mnt namespace indeed. For others, the stat() is called directly, thus the costs are not taken.

dialects/linux/dproc.c
```diff
@@ -1518,12 +1592,89 @@ process_proc_map(p, s, ss)
...
+		    int status;
+
+		    close(pipefd[1]);
+		    if (read(pipefd[0], &sv, sizeof(sv)) != sizeof(sv) ||
+			read(pipefd[0], &sb, sizeof(sb)) != sizeof(sb)) {
+			fprintf(stderr, "%s: can't read from pipe: %s\n",
+			    Pn, strerror(errno));
+			goto stat_directly;
+		    }
+
```
masatake, line 1660: Don't you need `close (pipefd[0])`?

masatake: I got
```
lsof: can't open pipes: Too many open files
lsof: can't open pipes: Too many open files
lsof: can't open pipes: Too many open files
lsof: can't open pipes: Too many open files
```

dialects/linux/dproc.c
masatake: > About fork and pipe costs, this patch mainly obtain correct stat result at the expense of these costs for the processes in different mnt namespace indeed. For others, the stat() is called directly, thus the costs are not taken.

(with your pull request + close(pipefs[0]) patch):
```
$ sudo time ./lsof > /dev/null
...(stderr)...
7.83user 318.46system 5:32.45elapsed 98%CPU (0avgtext+0avgdata 433404maxresident)k
8inputs+0outputs (0major+1565450minor)pagefaults 0swaps
$ git diff                    
diff --git a/dialects/linux/dproc.c b/dialects/linux/dproc.c
index bf6b216..8fb96e8 100644
--- a/dialects/linux/dproc.c
+++ b/dialects/linux/dproc.c
@@ -1658,6 +1658,7 @@ process_proc_map(p, s, ss)
                        goto stat_directly;
                    }
 
+                   close(pipefd[0]);
                    /* Wait the child process exit */
                    waitpid(pid, &status, 0);
```

Without your pull request:
```
$ git checkout master                                                     
Switched to branch 'master'                                                                                                                                                                   Your branch is up to date with 'origin/master'.                      
$ make -j 10
...
$ sudo time ./lsof > /dev/null
...(stderr)...
6.18user 7.35system 0:17.08elapsed 79%CPU (0avgtext+0avgdata 440348maxresident)k
0inputs+0outputs (0major+109423minor)pagefaults 0swaps
```

jiegec: That's a large performance regression. Can we compare the ns in a faster way? Or we will have to use the fallback approach.

jiegec: Can we move one process to another mount namespace and then move it back?

masatake:
```
$ uname -a
Linux living.shokupan 5.18.11-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Jul 12 22:52:35 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ sudo lsns -t mnt                                   
        NS TYPE NPROCS    PID USER            COMMAND
4026531841 mnt     596      1 root            /usr/lib/systemd/systemd --switched-root --system --deserialize 35
4026532188 mnt       1    754 root            ├─/usr/lib/systemd/systemd-udevd
4026532462 mnt       2    980 root            ├─/sbin/auditd
4026532464 mnt       1    978 systemd-oom     ├─/usr/lib/systemd/systemd-oomd
4026532465 mnt       1    985 systemd-resolve ├─/usr/lib/systemd/systemd-resolved
4026532475 mnt       2   1010 dbus            ├─/usr/bin/dbus-broker-launch --scope system --audit
4026532476 mnt       1   1018 root            ├─/usr/libexec/bluetooth/bluetoothd
4026532477 mnt       1   1061 chrony          ├─/usr/sbin/chronyd -F 2
4026532545 mnt       1   1053 root            ├─gpm -m /dev/input/mice -t exps2
4026532613 mnt       1   1026 root            ├─/usr/libexec/low-memory-monitor
4026532614 mnt       1   1036 root            ├─/usr/libexec/power-profiles-daemon
4026532682 mnt       1   1039 rtkit           ├─/usr/libexec/rtkit-daemon
4026532683 mnt       1   1040 root            ├─/usr/libexec/switcheroo-control
4026532686 mnt       1   1042 root            ├─/usr/lib/systemd/systemd-logind
4026532757 mnt       1   1106 root            ├─/usr/sbin/ModemManager
4026532758 mnt       1   1115 root            ├─/usr/sbin/NetworkManager --no-daemon
4026532759 mnt       6   1137 root            ├─php-fpm: master process (/etc/php-fpm.conf)
4026532784 mnt       5   1282 root            ├─/usr/sbin/httpd -DFOREGROUND
4026532786 mnt       1   2920 root            ├─/usr/libexec/upowerd
4026532790 mnt       1   4569 colord          └─/usr/libexec/colord
4026531862 mnt       1     74 root            kdevtmpfs
4026534282 mnt       1 208843 jet             catatonit -P
4026535188 mnt       1 209112 root            /bin/bash
```

masatake: It seems that fork is much more expensive than stat.

masatake: With this pull request:
```
$ sudo strace -f -c -e clone ./lsof > /dev/null
...
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
100.00  123.704179        5040     24540           clone
------ ----------- ----------- --------- --------- ----------------
100.00  123.704179        5040     24540           total
```
Improving the performance of clone() is good for the human being :-P
However, I guess there is no space for improvement.

You created a process for calling stat in a namespace repeatedly.
For a namespace, we need only one process running background.
For N namespace, we need only N processes. Running the process as a server.

dialects/linux/dproc.c
Hunter1016: It's my carelessless, thanks a lot.

BenBE:
Wouldn't it suffice to fork into each namespace just once and reuse that child from then on? This still gives a worst case of spawning a new process for each FD if they all happen to be in different namespaces, but for the common case this will maybe fork just a few processes handling all the information retrieval.

masatake:
"reuse" is the word I could not find. I wanted to say "make a process for each mnt namespace and reuse it".

masatake:
I'm looking for a reproducer that makes the situation this pull request is needed.

In terminal A:
```
[layer0]$ sudo unshare --mount --propagation private /bin/bash
[layer0]# PS1='[in ns]# '
[in ns]# dd if=/dev/zero of=xfs.img bs=40MB count=1
1+0 records in
1+0 records out
40000000 bytes (40 MB, 38 MiB) copied, 0.0319639 s, 1.3 GB/s
[in ns]# mkfs.xfs xfs.img 
meta-data=xfs.img                isize=512    agcount=2, agsize=4883 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1    bigtime=0 inobtcount=0
data     =                       bsize=4096   blocks=9765, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=1368, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0
[in ns]# mkdir /tmp/TEST
[in ns]# mount --make-private /tmp/TEST
mount: /tmp/TEST: not mount point or bad option.
       dmesg(1) may have more information after failed mount system call.
[in ns]# mount --make-private xfs.img /tmp/TEST
[in ns]# cat > /tmp/TEST/X &
[1] 1075968
[in ns]# ls -l /proc/1075968/fd
total 0
lrwx------. 1 root root 64 Jul 27 09:56 0 -> /dev/pts/7
l-wx------. 1 root root 64 Jul 27 09:56 1 -> /tmp/TEST/X
lrwx------. 1 root root 64 Jul 27 09:56 2 -> /dev/pts/7
```

Opening a new terminal (terminal B):
```
[layer0]$ sudo ls -l /tmp/TEST                                                                  
total 0                                                                                        
[layer0]$ sudo mount | grep /tmp/TEST      
[layer0]$ 
```
/tmp/TEST/X is not visible.

terminal B:
```
[layer0]$ sudo lsof -p 1075968
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1001/gvfs
      Output information may be incomplete.
lsof: WARNING: can't stat() fuse.portal file system /run/user/1001/doc
      Output information may be incomplete.
lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1000/gvfs
      Output information may be incomplete.
lsof: WARNING: can't stat() fuse.portal file system /run/user/1000/doc
      Output information may be incomplete.
COMMAND     PID USER   FD   TYPE DEVICE  SIZE/OFF      NODE NAME
cat     1075968 root  cwd    DIR  253,2      8192       103 /home/jet
cat     1075968 root  rtd    DIR  253,0      4096        96 /
cat     1075968 root  txt    REG  253,0     37568 134680698 /usr/bin/cat
cat     1075968 root  mem    REG  253,0 111950656 201538129 /usr/lib/locale/locale-archive
cat     1075968 root  mem    REG  253,0   2375712 134297710 /usr/lib64/libc.so.6
cat     1075968 root  mem    REG  253,0     26988     51241 /usr/lib64/gconv/gconv-modules.cache
cat     1075968 root  mem    REG  253,0    958728 134295823 /usr/lib64/ld-linux-x86-64.so.2
cat     1075968 root  mem    REG  253,0      3284   1030507 /usr/lib/locale/en_US.utf8/LC_TIME
cat     1075968 root    0u   CHR  136,7       0t0        10 /dev/pts/7
cat     1075968 root    1w   REG    7,1         0       131 /tmp/TEST/X
cat     1075968 root    2u   CHR  136,7       0t0        10 /dev/pts/7
```

lsof reports /tmp/TEST/X. This pull request is not applied to this lsof binary.
This lsof binary is one shipped as part of Fedora36.

I wonder when this patch is needed.
It will be nice to have a reproducer implemented with unshare command.

Hunter1016:
> I'm looking for a reproducer that makes the situation this pull request is needed.
> 
> In terminal A:
> 
> ```
> [layer0]$ sudo unshare --mount --propagation private /bin/bash
> [layer0]# PS1='[in ns]# '
> [in ns]# dd if=/dev/zero of=xfs.img bs=40MB count=1
> 1+0 records in
> 1+0 records out
> 40000000 bytes (40 MB, 38 MiB) copied, 0.0319639 s, 1.3 GB/s
> [in ns]# mkfs.xfs xfs.img 
> meta-data=xfs.img                isize=512    agcount=2, agsize=4883 blks
>          =                       sectsz=512   attr=2, projid32bit=1
>          =                       crc=1        finobt=1, sparse=1, rmapbt=0
>          =                       reflink=1    bigtime=0 inobtcount=0
> data     =                       bsize=4096   blocks=9765, imaxpct=25
>          =                       sunit=0      swidth=0 blks
> naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
> log      =internal log           bsize=4096   blocks=1368, version=2
>          =                       sectsz=512   sunit=0 blks, lazy-count=1
> realtime =none                   extsz=4096   blocks=0, rtextents=0
> [in ns]# mkdir /tmp/TEST
> [in ns]# mount --make-private /tmp/TEST
> mount: /tmp/TEST: not mount point or bad option.
>        dmesg(1) may have more information after failed mount system call.
> [in ns]# mount --make-private xfs.img /tmp/TEST
> [in ns]# cat > /tmp/TEST/X &
> [1] 1075968
> [in ns]# ls -l /proc/1075968/fd
> total 0
> lrwx------. 1 root root 64 Jul 27 09:56 0 -> /dev/pts/7
> l-wx------. 1 root root 64 Jul 27 09:56 1 -> /tmp/TEST/X
> lrwx------. 1 root root 64 Jul 27 09:56 2 -> /dev/pts/7
> ```
> 
> Opening a new terminal (terminal B):
> 
> ```
> [layer0]$ sudo ls -l /tmp/TEST                                                                  
> total 0                                                                                        
> [layer0]$ sudo mount | grep /tmp/TEST      
> [layer0]$ 
> ```
> 
> /tmp/TEST/X is not visible.
> 
> terminal B:
> 
> ```
> [layer0]$ sudo lsof -p 1075968
> lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1001/gvfs
>       Output information may be incomplete.
> lsof: WARNING: can't stat() fuse.portal file system /run/user/1001/doc
>       Output information may be incomplete.
> lsof: WARNING: can't stat() fuse.gvfsd-fuse file system /run/user/1000/gvfs
>       Output information may be incomplete.
> lsof: WARNING: can't stat() fuse.portal file system /run/user/1000/doc
>       Output information may be incomplete.
> COMMAND     PID USER   FD   TYPE DEVICE  SIZE/OFF      NODE NAME
> cat     1075968 root  cwd    DIR  253,2      8192       103 /home/jet
> cat     1075968 root  rtd    DIR  253,0      4096        96 /
> cat     1075968 root  txt    REG  253,0     37568 134680698 /usr/bin/cat
> cat     1075968 root  mem    REG  253,0 111950656 201538129 /usr/lib/locale/locale-archive
> cat     1075968 root  mem    REG  253,0   2375712 134297710 /usr/lib64/libc.so.6
> cat     1075968 root  mem    REG  253,0     26988     51241 /usr/lib64/gconv/gconv-modules.cache
> cat     1075968 root  mem    REG  253,0    958728 134295823 /usr/lib64/ld-linux-x86-64.so.2
> cat     1075968 root  mem    REG  253,0      3284   1030507 /usr/lib/locale/en_US.utf8/LC_TIME
> cat     1075968 root    0u   CHR  136,7       0t0        10 /dev/pts/7
> cat     1075968 root    1w   REG    7,1         0       131 /tmp/TEST/X
> cat     1075968 root    2u   CHR  136,7       0t0        10 /dev/pts/7
> ```
> 
> lsof reports /tmp/TEST/X. This pull request is not applied to this lsof binary. This lsof binary is one shipped as part of Fedora36.
> 
> I wonder when this patch is needed. It will be nice to have a reproducer implemented with unshare command.

This patch is mainly for memory-mapped files, i.e. 'mem' in the 'FD' column. I'm afraid the situation above is not related to what this patch handles.

masatake:
> This patch is mainly for memory-mapped files, i.e. 'mem' in the 'FD' column. I'm afraid the situation above is not related to what this patch handles. @masatake

Oh, I see.
I can use https://github.com/masatake/mmap for reproducing.

I think setns is not needed. Did you try /proc/$pid/map_files/*?
lsfd works well without using setns.

Hunter1016:
> > This patch is mainly for memory-mapped files, i.e. 'mem' in the 'FD' column. I'm afraid the situation above is not related to what this patch handles. @masatake
> 
> Oh, I see. I can use https://github.com/masatake/mmap for reproducing.
> 
> I think setns is not needed. Did you try /proc/$pid/map_files/*? lsfd works well without using setns.

It's a little confusing, how to acquire 'inode, major, minor', from link files under  /proc/$pid/map_files/* ?

masatake:
```
[layer0]$ sudo stat -L /proc/1/map_files/7f44c3452000-7f44c3454000
  File: /proc/1/map_files/7f44c3452000-7f44c3454000
  Size: 958728          Blocks: 1880       IO Block: 4096   regular file
Device: 253,0   Inode: 134295823   Links: 1
Access: (0755/-rwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:ld_so_t:s0
Access: 2022-07-26 18:55:51.451937735 +0900
Modify: 2022-07-05 20:59:06.000000000 +0900
Change: 2022-07-18 18:46:36.208219856 +0900
 Birth: 2022-07-18 18:46:36.142219642 +0900
                                                                                                                                               
```

`stat -L` will work even if the target of the symlink is not visible from the current mnt spae.

Have you ever got a question about the mechanism that lsof can print the information about the deleted file? SNS, super-natural symlinks are behind:-P

masatake:
Reproduced.
As I wrote I used my mmap command (https://github.com/masatake/mmap).

The terminal A:
```
[in ns]# git clone https://github.com/masatake/mmap
Cloning into 'mmap'...
remote: Enumerating objects: 29, done.
remote: Counting objects: 100% (1/1), done.
remote: Total 29 (delta 0), reused 0 (delta 0), pack-reused 28
Receiving objects: 100% (29/29), 7.64 KiB | 3.82 MiB/s, done.
Resolving deltas: 100% (8/8), done.
[in ns]# cd mmap
[in ns]# gcc -o mmap ./mmap.c 
[in ns]# dd if=/dev/zero of=/tmp/TEST/MMAP bs=2MB count=1
1+0 records in
1+0 records out
2000000 bytes (2.0 MB, 1.9 MiB) copied, 0.00187238 s, 1.1 GB/s
[in ns]# ./mmap -m 1  --protection r--p -f /tmp/TEST/MMAP 
pid: 1120741
addr: 0x7faeb0300000

```

The mmap command keeps the mapping.

The terminal B:
```
[layer0]$ sudo lsof -p 1120741  | grep MMAP 
...
mmap    1120741 root  mem    REG    7,1                132 /tmp/TEST/MMAP (stat: No such file or directory)
mmap    1120741 root    3r   REG    7,1  2000000       132 /tmp/TEST/MMAP
[layer0]$ sudo ./lsof -p 1120741  | grep MMAP
...
mmap    1120741 root  mem    REG    7,1  2000000       132 /tmp/TEST/MMAP
mmap    1120741 root    3r   REG    7,1  2000000       132 /tmp/TEST/MMAP
```

Your changes was applied to ./lsof.
Works fine.


The LATEST lsfd uses `map_files`.
```
[layer0]$ sudo ~/var/util-linux/lsfd -p 1120741  | grep MMAP
mmap    1120741 root    mem  r--  REG  loop1     0        132 /tmp/TEST/MMAP
mmap    1120741 root      3  r--  REG  loop1  2185        132 /tmp/TEST/MMAP
[layer0]$ ls -l /dev/loop1
brw-rw----. 1 root disk 7, 1 Jul 27 09:54 /dev/loop1
```
lsfd decodes 7,1 and prints it as "loop1". lsfd gets 7,1 without doing fork&stat.

Hunter1016:
> Reproduced. As I wrote I used my mmap command (https://github.com/masatake/mmap).
> 
> The terminal A:
> 
> ```
> [in ns]# git clone https://github.com/masatake/mmap
> Cloning into 'mmap'...
> remote: Enumerating objects: 29, done.
> remote: Counting objects: 100% (1/1), done.
> remote: Total 29 (delta 0), reused 0 (delta 0), pack-reused 28
> Receiving objects: 100% (29/29), 7.64 KiB | 3.82 MiB/s, done.
> Resolving deltas: 100% (8/8), done.
> [in ns]# cd mmap
> [in ns]# gcc -o mmap ./mmap.c 
> [in ns]# dd if=/dev/zero of=/tmp/TEST/MMAP bs=2MB count=1
> 1+0 records in
> 1+0 records out
> 2000000 bytes (2.0 MB, 1.9 MiB) copied, 0.00187238 s, 1.1 GB/s
> [in ns]# ./mmap -m 1  --protection r--p -f /tmp/TEST/MMAP 
> pid: 1120741
> addr: 0x7faeb0300000
> ```
> 
> The mmap command keeps the mapping.
> 
> The terminal B:
> 
> ```
> [layer0]$ sudo lsof -p 1120741  | grep MMAP 
> ...
> mmap    1120741 root  mem    REG    7,1                132 /tmp/TEST/MMAP (stat: No such file or directory)
> mmap    1120741 root    3r   REG    7,1  2000000       132 /tmp/TEST/MMAP
> [layer0]$ sudo ./lsof -p 1120741  | grep MMAP
> ...
> mmap    1120741 root  mem    REG    7,1  2000000       132 /tmp/TEST/MMAP
> mmap    1120741 root    3r   REG    7,1  2000000       132 /tmp/TEST/MMAP
> ```
> 
> Your changes was applied to ./lsof. Works fine.
> 
> The LATEST lsfd uses `map_files`.
> 
> ```
> [layer0]$ sudo ~/var/util-linux/lsfd -p 1120741  | grep MMAP
> mmap    1120741 root    mem  r--  REG  loop1     0        132 /tmp/TEST/MMAP
> mmap    1120741 root      3  r--  REG  loop1  2185        132 /tmp/TEST/MMAP
> [layer0]$ ls -l /dev/loop1
> brw-rw----. 1 root disk 7, 1 Jul 27 09:54 /dev/loop1
> ```
> 
> lsfd decodes 7,1 and prints it as "loop1". lsfd gets 7,1 without doing fork&stat.

Helpful illustration, and I've experienced the performance regression already. I'll try to optimize this by means of reading /proc/$pid/map_files/*.

Hunter1016:
Forced-updated, by accessing symlinks under /proc/[pid]/map_files/, it's quite straight-forward than the previous way, and do not come with much performance regression. Please kindly review.@masatake

dialects/linux/dproc.c
```diff
@@ -1395,6 +1398,31 @@ process_id(idp, idpl, cmd, uid, pid, ppid, pgid, tid, tcmd)
 	return(0);
 }
 
+/* compare mount namespace of this lsof proces and the target process */
```
BenBE, line 1401:
```suggestion
/* compare mount namespace of this lsof process and the target process */
```

dialects/linux/dproc.c
```diff
@@ -1518,11 +1552,44 @@ process_proc_map(p, s, ss)
 		efs = sv = 1;
 	    else
 		efs = 0;
-	    if (!efs) {
-		if (HasNFS)
-		    sv = statsafely(fp[6], &sb);
-		else
-		    sv = stat(fp[6], &sb);
+
+	    /* For prcesses in different mount namespace from lsof process,
```
BenBE, line 1556:
```suggestion
	    /* For processes in different mount namespace from lsof process,
```

BenBE, line 1556:
```suggestion
	    /* For processes in different mount namespace from lsof process,
```

BenBE (commented):
Some minor typos …

dialects/linux/dproc.c
```diff
@@ -1518,11 +1552,44 @@ process_proc_map(p, s, ss)
...
+	    if (diff_mntns) {
+		char path[MAP_PATH_LENGTH];
+		char addr[ADDR_LENGTH];
+		uint64_t start, end;
+		int ret;
+
+		if (sscanf(fp[0], "%lx-%lx", &start, &end) != 2)
+		    goto stat_directly;
+
+		ret = snprintf(addr, sizeof(addr), "%lx-%lx", start, end);
```
jiegec, line 1569: Why is it need to sscanf() and then snprintf()?

Hunter1016: Format of files under /proc/[pid]/maps and /proc/[pid]/map_files/* are a little different, mainly on padding zeros.

dialects/linux/dproc.c
BenBE: @Hunter1016 I think you missed this one …

Hunter1016: Ah, I've forced-updated. Thanks a lot. @BenBE

dialects/linux/dproc.c
jiegec: Thanks for clarification.

jiegec:
@Hunter1016 I am fine with the changes. Could you add a test case to test the behavior with/without this pr and add an changelog entry to `00DIST`?

Hunter1016:
> 



> @Hunter1016 I am fine with the changes. Could you add a test case to test the behavior with/without this pr and add an changelog entry to `00DIST`?

I've completed the test case, i.e. dialects/linux/tests/case-20-mmap.bash, which use `ps` command to identify the mmap process in the specified namespace, but the test report shows `ps: command not found`. I think it's because we run the tests in a container environment, may lack of this command, is it possible to add it? @jiegec

masatake:
Could you add `procps` to the dnf of stream8 (https://github.com/lsof-org/lsof/blob/bcc125b7c2a160da019dd8089b999bc6d3486616/.circleci/config.yml#L43)? This addition may be part of [403c39f](https://github.com/lsof-org/lsof/pull/239/commits/403c39fab58a3e35ba3caaebb5b93743d91a3324).

Hunter1016:
> 



> Could you add `procps` to the dnf of stream8 (
> 
> https://github.com/lsof-org/lsof/blob/bcc125b7c2a160da019dd8089b999bc6d3486616/.circleci/config.yml#L43
> 
> )? This addition may be part of [403c39f](https://github.com/lsof-org/lsof/pull/239/commits/403c39fab58a3e35ba3caaebb5b93743d91a3324).
> This is applicable to fedora36, too.

I've added this in an individual commit to avoid confusion.

Hunter1016:
Changes on 00DIST and test case, which use `unshare` command to create a new namespace, are provided. Please kindly review. @jiegec

masatake:
@jiegec Could you update 00CREDITS?

jiegec:
@Hunter1016 Which one do you prefer to have in `00CREDITS`?

1. @Hunter1016 
2. huteng.ht (@Hunter1016)
3. You Real Name Here (@Hunter1016)

PR #240, opened 2022-08-09
docs: add a simple example to the FAQ

chapmanjacobd:
It might be somewhat obvious but I keep having to relearn it

### DIFF
--- a/00CREDITS
+++ b/00CREDITS
@@ -567,8 +567,9 @@ provided test systems where I was able to do development work.
 	Robert Geislinger (Alienmaster)
 	Subhaditya Nath
 	Sergei Trofimovich (@trofi)
-	Danilo Spinella (@danyspin97) and
-	Jiajie Chen (@jiegec)
+	Danilo Spinella (@danyspin97)
+	Jiajie Chen (@jiegec) and
+	Jacob Chapman (@chapmanjacobd)
 
 If I have omitted a contributor's name, the fault is wholly mine,
 and I apologize for the error.
--- a/00DIST
+++ b/00DIST
@@ -5375,5 +5375,8 @@ July 14, 2018
 		[linux] obtain correct information of memory-mapped file.
 		Provided by Teng Hu in #239
 
+		Updated 00FAQ with lookup to open files via mountpoint
+		Provided by Jacob Chapman in #240
+
 The lsof-org team at GitHub
-April 28, 2022
+August 10, 2022


jiegec:
Please squash your commits into one and format the commit message.

chapmanjacobd:
okay :heavy_check_mark:

00FAQ
```diff
@@ -2203,8 +2203,15 @@ ______________________________________________________________________
...
-	of those files, or use the lsof +D and +d options.
+	directory, you need to either specify the individual path names
+	of those files, their common mount point, or use the lsof +D 
+	and +d options.
+	
+	For example, if you wanted to check the files open in your /lsof 
+	folder then the following might be the quickest way to get
+	the answer you are looking for:
+
+	lsof / | grep /lsof | grep -v DIR
```
jiegec, line 2214: Why not `lsof /lsof`?

chapmanjacobd: This section of the FAQ is 

```
3.21.4	Why won't lsof find all the open files in a directory?
```

`lsof /lsof` will give different results from `lsof / | grep /lsof` unless `/lsof` is a mountpoint. 

You would need something like `lsof +D /lsof` to be without `grep` or maybe there is another feature of lsof that I am unaware of which would be similar

jiegec: I see.

jiegec (approved):
Please add yourself to `00CREDITS` and change log to `00DIST` in a separate commit.

00CREDITS
```diff
@@ -569,6 +569,7 @@ provided test systems where I was able to do development work.
 	Sergei Trofimovich (@trofi)
 	Danilo Spinella (@danyspin97) and
 	Jiajie Chen (@jiegec)
+	Jacob Chapman (@chapmanjacobd)
```
jiegec, line 572: Please move the `and` to Line 571

jiegec: Please format the last commit message as well.

chapmanjacobd: nobody else has an `and`. it likely makes more sense to remove that or place both names on one line. I'm not 100% sure what you want to do

either way I feel like that change is out of scope for this PR. I never modified line 570 or 571. if a previous PR was incorrect it is the maintainer's job to fix it. feel free to add commits to this branch

jiegec: I mean you should change it to:

```
trofi
danyspin97
jiegec and
chapmanjacobd
```

chapmanjacobd: mmmmk
I thought it was a line delimited list

jiegec:
Merged in commit 7a7d1827faaf5fe558f6c0e16861f683b111fd97

PR #241, opened 2022-08-17
freebsd: net/raw_cb.h no longer exists

kprovost:
As of f63cb32c1988561136fabdcc54d16cd200b666d9 (in FreeBSD/src) the net/raw_cb.h
header no longer exists. It was removed as part of the retirement of
4.4BSD raw sockets.

Happily this header is not actually required for lsof, so we can just
not include it to fix the ports build.

See also: https://reviews.freebsd.org/D36240
Sponsored by:   Rubicon Communications, LLC ("Netgate")


masatake:
@lrosenman, please update 00CREDITS.

Issue #187, opened 2022-01-19
warning message not supressed in AIX


sanumesh:
Hello

In AIX, we still see the warning message related to Kernel ID check in spite of using the -w option

WARNING: compiled for AIX version 6.1.0.0; this is 7.2.0.0.

Solution :
In dialects/aix/dproc.c
Function : void ckkv(d, er, ev, ea)

The if (Fwarn) check is missing for AIX levels above 5.0.0.0

 if	AIXV<5000
...
	if (Fwarn)
	    return;
 else
...
endif

masatake:
A pull request for the solution is welcome.

masatake:
@sanumesh 
> Solution :
...

A pull request is welcome.

PR #244, opened 2022-08-24
Test on all AIX versions to solve issue #187

oldmanhere:
IBM lsof was compiled on AIX 6.1, and targets are AIX 6.1, 7.1 and 7.2.
On AIX 7.1, we have "lsof: WARNING: compiled for AIX version 6.1.0.0; this is 7.1.0.0.", even if all is ok.

This test, for old AIX versions under 5.0 on lines 245 and 246,  need to be done everywhere.

I don't have compilers and so on to test that new code, but could be ok if your compiler accept it.

### DIFF
--- a/dialects/aix/dproc.c
+++ b/dialects/aix/dproc.c
@@ -313,6 +313,9 @@ ckkv(d, er, ev, ea)
 	char buf[64];
 	struct utsname u;
 
+	if (Fwarn)
+	    return;
+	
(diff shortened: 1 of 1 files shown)


masatake:
The commit log is too short for the change.
Could you write what you wrote in the initial comment of this pull request to the commit log?

Could you update 00DIST?

oldmanhere:
@masatake I don't have any idea how to accomplish that in the github UI. Sorry.

PR #248, opened 2022-09-12
Configure: fix configure tests broken with Clang 15 (-Wimplicit-int)

thesamesam:
Clang 15 makes -Wimplicit-int an error by default.

Before this fix, configure would think localtime() and strftime() support was not present.

Signed-off-by: Sam James <sam@gentoo.org>


thesamesam:
Thanks!

PR #250, opened 2022-09-16
    All supported FreeBSD releases have dup2 and closefrom

emaste:


emaste:
Likely best to merge #196 before this one to make it clear that older version is unsupported.

PR #252, opened 2022-09-16
Add FreeBSD 14 snapshots to CI matrix

emaste:


bsdimp:
Consider this patch
```
diff --git a/dialects/freebsd/dlsof.h b/dialects/freebsd/dlsof.h
index 8a43d79..4d9dd0f 100644
--- a/dialects/freebsd/dlsof.h
+++ b/dialects/freebsd/dlsof.h
@@ -315,7 +315,11 @@ int     open(const char *, int, ...);
 #define        vasprintf vasprintf_kernel_lsof
 #define        uintfptr_t      int
 #define        _SYS_LIBKERN_H_
+#define tick_sbt 1
+#define pause kernel_pause
 #include <sys/file.h>
+#undef pause
+#undef tick_sbt

 /*
  * Attempt to remove the circumventions.
```

PR #262, opened 2022-10-11
tests: removing redundant statement

Sigmw:


jiegec:
Why is it redundant?

Sigmw:
> Why is it redundant?

Sorry, not redundant, just a repeated statement in if conditions.

Issue #267, opened 2022-11-22
[FEATURE] Build system rewrite in autotools or meson


thesamesam:
**Is your feature request related to a problem? Please describe.**

lsof's current `Configure` script is handwritten and causes problems for cross-compilation for example as it tries to build-and-execute binaries to determine properties of the system.

This requires awkward patching and workarounds downstream in e.g. Gentoo.

**Describe the solution you'd like**

A rewrite of the build system to either autotools or meson, both of which would solve this problem and are standard in the community.

**Describe alternatives you've considered**

The configure script could develop additional options to allow overrides for e.g. cross-compile, but this wouldn't solve all the issues (see below).

**Additional context**

Various cross bugs which are awkward to handle/patch correctly:
- https://bugs.gentoo.org/432120
- https://bugs.gentoo.org/546636

The handwritten configure script often doesn't conform to modern standards:
- https://github.com/lsof-org/lsof/pull/265
- https://github.com/lsof-org/lsof/pull/248

thesamesam:
@jiegec @lrosenman if you're fine with this in principle, we (Gentoo) are happy to work on this and provide a PR, but didn't want to waste effort on it if it would be rejected. Let us know what you think, thanks!

lrosenman:
I'm fine with it.  Can you investigate what changes would also be needed for the FreeBSD sysutils/lsof port for whatever you do?

SoapGentoo:
Given that the Autotools are UNIX-agnostic, I don't see any major hurdles for the FreeBSD port. Do you prefer Meson or Autotools? Here's a few points to consider:

1. Autotools is acceptable for FreeBSD for base system packages, whereas Meson isn't (due to it being written in python)
2. Autotools generally only requires POSIX tools (sed, awk, sh, a C compiler), whereas Meson requires python and ninja.

lrosenman:
FreeBSD ports supports autotools natively with the framework.  

we can talk after you present a fix

eli-schwartz:
Autotools is a very defensible choice, particularly if a project only needs Unix support. It helps a *lot* if the project doesn't need to use libtool for great slowness (and lsof doesn't have any libraries, the one that it does use is an internal implementation detail and easily avoided).

That being said, https://sr.ht/~lattis/muon/ does exist to build meson projects without python.

jiegec:
I have a WIP autotools-based build system rewrite in branch `autotools`.

thesamesam:
[...]

PR #270, opened 2023-01-13
Autotools build system

jiegec:
Migrate to autotools build system #267. The old build system will not be deleted so autotools should generate the same cflags/ldflags as the old one.

- [x] Building on Linux Debian bookworm
- [x] Building on Darwin 22
- [x] Building on FreeBSD 13
- [x] Testing
- [x] Installation
- [x] Manpages
- [x] CI
- [x] Documentation


jiegec:
Close #267.

PR #277, opened 2023-01-28
[linux] Do not hard-code fd numbers in epoll test

subnut:
In case file descriptor 3 and 4 are being used already, the executable created by epoll.c exits with exit code 1, and prints the message -

	epoll_ctl<6>: Invalid argument

causing case-20-epoll.bash to FAIL.


jiegec:
Nice job, could you provide information to reproduce the issue? e.g. Linux distro.

subnut:
Void Linux's `xbps-src` build system

PR #278, opened 2023-01-29
arg: Remove unnecessary initialization assignments

likunyur:
The dn pointer variable will be assigned before use. It does not need to initialize the assignment.

Signed-off-by: Li kunyu <kunyu@nfschina.com>


jiegec:
I agree that the initialization is redundant in the specific case, but it is also good to initialize everything to invalid before usage. I usually ignore those warnings report by analyzers.

PR #279, opened 2023-02-07
[WIP] Expose functionality as a re-usable library

jiegec:
This pr implements #268 by splitting code into two parts: liblsof(the library) and lsof(the cli). The pr is unfinished, but is half way done. It is a VERY BIG pull request and I try my best to maintain the lsof cli behavior. I am opening this pr to request for more comments. Progress:

- [x] Re-arrange directory structure, build liblsof using libtool.
- [x] Add initial library interface with Doxygen.
- [x] Move global and static variables to lsof context.
- [x] Rewrite LTbasic & LTszoff with liblsof.
- [x] Set visibility to hidden for internal functions of liblsof.
- [x] Migrate to Linux/Darwin/FreeBSD/NetBSD/OpenBSD/Solaris/OpenIndiana. Support for AIX/HP-UX/SCO/UnixWare will be broken, but I do not have those environments to fix.
- [x] Write some utilities with liblsof as an example, maybe re-implement a netstat?
- [x] Selection result reporting, implemented as an array of tagged union, e.g. selection by pid success, selection by command name failed etc.
- [x] Add optional fields to `struct lsof_file`, e.g. network address, TCP state.
- [ ] Allow liblsof to read info when running as non-root on NetBSD and Solaris by switching to /proc-based.
- [ ] Replace all exit calls in liblsof with proper error reporting.
- [ ] Migrate more options from cli to liblsof until cli only uses public interface of liblsof.
- [ ] Fix memory leaks.

I'd love to see more comments on the API interface (You can find some at #268).

PR #285, opened 2023-04-03
Expose functionality as a re-usable library - Part 3

jiegec:
Part 3 of #279: 

1. do not check command flags in dialect-specific code, always save the information into `Lf`.
2. Convert `Lf->access` to enum

### DIFF
--- a/lib/print.c
+++ b/lib/print.c
@@ -36,9 +36,9 @@
  */
 
 /*
- * print_access() - print enum lsof_file_access_mode
+ * access_to_char() - convert enum lsof_file_access_mode to char
  */
-char print_access(enum lsof_file_access_mode access) {
+char access_to_char(enum lsof_file_access_mode access) {
     switch (access) {
     default:
     case LSOF_FILE_ACCESS_NONE:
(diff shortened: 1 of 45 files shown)


lib/print.c
```diff
@@ -0,0 +1,53 @@
...
+#include "lsof.h"
+
+/*
+ * Local definitions, structures and function prototypes
+ */
+
+/*
+ * print_access() - print enum lsof_file_access_mode
+ */
+char print_access(enum lsof_file_access_mode access) {
```
BenBE, line 41: The naming for `print_access` is a bit misleading as this function determines the character used to indicate the kind of access on the file; it doesn't actually print anything.

Also it might be worth distinguishing the `LSOF_FILE_ACCESS_NONE` case and undefined by returning `?` for the default case. This should not happen, but better provide this distinction then silently drop it …

BenBE (approved):
Two small suggestions, nothing big.

PR LGTM; didn't check the actual removed block braces to be the right ones due to all indentation changes. The pairing at least looks consistent.

lib/print.c
jiegec: The space is expected to fill one character's width in the tabular output.

PR #290, opened 2023-05-12
Library part5

jiegec:
Part 5 of pr #279

- Move global variable to context
- Add lsof APIs:
  - lsof_get_api_version
  - lsof_get_library_version
  - lsof_set_output_stream
  - lsof_avoid_blocking
  - lsof_avoid_forking
  - lsof_logic_and
  - lsof_select_process

### DIFF
--- a/lib/common.h
+++ b/lib/common.h
@@ -253,7 +253,7 @@ struct l_dev {
 #    include "dlsof.h"
 
 #    include <sys/types.h> /* just in case -- because utmp.h
-					 * may need it */
+                     * may need it */
 #    include <regex.h>
 
 #    if defined(EMPTY)
@@ -991,9 +991,6 @@ extern struct mounts *Mtprocfs;
 extern int Mxpgid;
 extern int Mxpid;
 extern int Mxuid;
-extern gid_t Mygid;
-extern int Mypid;
-extern uid_t Myuid;
 extern int Ndev;
 
 #    if defined(HASNLIST)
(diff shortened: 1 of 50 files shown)


lib/common.h
```diff
@@ -1120,18 +1028,484 @@ extern int UdpStXn;
...
+#    define HbyFrd (ctx->sfile_hash_file_raw_device)
+#    define HbyFrdCt (ctx->sfile_hash_file_raw_device_count)
+#    define HbyFsd (ctx->sfile_hash_file_system)
+#    define HbyFsdCt (ctx->sfile_hash_file_system_count)
+#    define HbyNm (ctx->sfile_hash_name)
+#    define HbyNmCt (ctx->sfile_hash_name_count)
+#    define HbyCd (ctx->sfile_hash_clone)
+#    define HbyCdCt (ctx->sfile_hash_clone_count)
+/* solaris zone */
+#    define ZoneArg (ctx->sel_zone)
```
BenBE, line 1498: As there are quite a lot of these convenience macros they run the risk of overly polluting the namespace if not needed. Please split into a separate header (if publicly exposed).

BenBE (commented):
LGTM. Some minor code style things re alignment of line continuations left.

lib/common.h
jiegec: This is actually a private header for liblsof. The public interface is `lsof.h`.

PR #316, opened 2024-05-15
[linux] closefrom_shim: Add optimized fallback for platforms without closefrom or close_range

MarcT512:
Add an optimized fallback in closefrom_shim() for Linux, if the platform doesn't support closefrom or close_range.
Iterate /proc/self/fd and close anything which is not that directory itself.

This can save significant time if "ulimit -n" (max open files) is very high. For example, in some Docker containers the limit can be 2^30 (~ 1 billion), and lsof may take 10-30 minutes to start up.


jiegec:
Interesting optimization, I hope to see some numbers, if possible. But I think it only applies on linux distributions with older Linux kernel versions, like 2.x?

PR #321, opened 2024-07-29
Remove myself from the team of linux dialect

masatake:
People can be confused with my name on the list of maintainers.


jiegec:
Thanks for your hard work ;)

PR #323, opened 2024-10-05
configure.ac: get version from version file.

Z572:
see https://stackoverflow.com/questions/8559456/read-a-version-number-from-a-file-in-configure-ac


jiegec:
Why?

PR #337, opened 2025-09-08
[freebsd] Update supported FreeBSD versions

yasuhirokimura:
* Add FreeBSD 16
* Remove FreeBSD 12 (EoL on December 31, 2023)

### DIFF
--- a/Configure
+++ b/Configure
@@ -1069,11 +1069,6 @@ case $LSOF_TGT in	# {
     # If the FreeBSD version isn't pre-defined, determine it.
 
       case $LSOF_VSTR in	# {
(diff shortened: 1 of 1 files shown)


yasuhirokimura:
> Thanks for you contributions! We are having some problems in FreeBSD CI, can you verify that it works on FreeBSD 16?

I confirmed build succeeds with FreeBSD 16.0-CURRENT amd64 with following steps.

1.  `env FREEBSD_SYS=/usr/src/sys LSOF_OPINC="-idirafter /usr/src/sys" bash Configure -n freebsd`
2.  `make`

Issue #311, opened 2024-03-22
[BUG] fd numbers also truncated in -F fd *output for other programs*


stephane-chazelas:
**Describe the bug**

I can (sort of) understand the truncation of the fd number (and asterisk alerting about it) in:

```
$ bash -c 'ulimit -n 20000; exec 12345<> /dev/null; lsof -w -ap "$$" /dev/null; exit'
COMMAND  PID     USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
bash    9385 chazelas *345u   CHR    1,3      0t0    5 /dev/null
```

Where the FD column has a fixed limited width that hadn't accounted for fds one day possibly be greater than 999 decades ago

But IMO, it doesn't make sense in the *output for other programs* where there's no reason for the width to be limited:

```
$ bash -c 'ulimit -n 20000; exec 12345<> /dev/null; lsof -w -F fd -ap "$$" /dev/null; exit'
p9431
f*345
```

Expected:

```
$ bash -c 'ulimit -n 20000; exec 12345<> /dev/null; lsof -w -F fd -ap "$$" /dev/null; exit'
p9431
f12345
```

(reproduced with lsof from Ubuntu 22.04 and current git HEAD (e30113e5eb151a83d7a4d0935f58b6b542ed31eb)).

PR #353, opened 2026-03-12
feat: add native JSON output (-J/-j flags)

kurok:
## Summary

- Add `-J` flag for nested JSON output (single JSON object with processes→files hierarchy)
- Add `-j` flag for flat JSON Lines output (one denormalized JSON object per open file, one per line)
- Both formats reuse the existing `-F` field selection mechanism for choosing which fields to include

## Motivation

The legacy `-F` field output uses single-character field IDs with newline/NUL terminators. While machine-parseable, it requires custom parsers. Native JSON enables direct ingestion into logging pipelines (Datadog, Splunk, Elastic Stack) and seamless parsing by Python, Go, Nushell, and `jq`.

## Changes

| File | Changes |
|------|---------|
| `src/store.c` | Add `Fjson`, `Fjsonl`, `Fjson_first_proc` global flags |
| `lib/common.h` | Extern declarations for new globals |
| `src/main.c` | Option parsing for `-J`/`-j`, mutual exclusivity validation, default field selection |
| `src/print.c` | JSON helpers (string escaping, field emitters), `json_print_file()`, `json_print_proc_fields()`, JSON branch in `print_proc()`, envelope functions |
| `lib/proto.h` | Declarations for `json_open_envelope`/`json_close_envelope` |
| `lib/dialects/*/Makefile` | Add `version.h` dependency for `print.o` |
| `Lsof.8` | Man page documentation for `-J` and `-j` |
| `tests/case-30-json-output.bash` | 6 test cases for `-J` nested JSON |
| `tests/case-31-jsonl-output.bash` | 3 test cases for `-j` JSON Lines |
| `Makefile.am` | Register new tests |

## Design decisions

- **No external dependencies** — hand-written JSON formatting (lsof data is simple enough)
- **Large numeric values as strings** — `size`, `offset`, `inode` emitted as JSON strings to avoid IEEE 754 precision loss
- **Numeric FDs as strings** — directly addresses issue #311 (fd truncation in field output)
- **TCP/TPI as nested object** — `tcp_info: {state, recv_queue, send_queue, ...}` in both formats
- **Fields with no value omitted** — not set to null

## Test plan

- [ ] `-J` produces valid JSON (parseable by `python3 -m json.tool`)
- [ ] `-j` produces valid JSON Lines (each line parseable)
- [ ] Field selection with `-Fpcfn` limits output to selected fields only
- [ ] `-J` and `-j` are mutually exclusive (error if both specified)
- [ ] `-J`/`-j` and `-t` are mutually exclusive
- [ ] Empty results produce valid JSON (`{"lsof_version":"...","processes":[]}` for `-J`, empty for `-j`)
- [ ] `+J`/`+j` produce error messages
- [ ] Clean build from scratch on Darwin


src/print.c
```diff
@@ -93,6 +95,83 @@ static char *lkup_svcnam(struct lsof_context *ctx, int h, int p, int pr,
...
+            break;
+        }
+        p++;
+    }
+}
+
+static void json_print_str(int *sep, const char *key, const char *val) {
+    if (*sep)
+        putchar(',');
+    printf("\"%s\":\"", key);
```
jiegec: The key might also require escape?

PR #368, opened 2026-08-11
[linux] fix segfault when eventpoll fdinfo is unavailable

daeho-ro:
`tfd_count` is declared `size_t` (`dproc.c:97`), but `process_id()` uses `-1` as its "no fdinfo" sentinel (`dproc.c:1222`). It becomes `SIZE_MAX`, so the `fi.tfd_count > 0` guard passes where the `!= -1` siblings skip, and `snp_eventpoll()` qsorts `SIZE_MAX` entries of a 32-entry array:

```
#4  __GI___qsort_r (total_elems=18446744073709551615, size=4, cmp=fd_compare)
#5  process_id ()
```

This hits any run with `OffType != OFFSET_FDINFO` that reports a process holding an epoll descriptor. Declaring it `int` matches `tfds[]` and the `eventfd_id`/`pid` sentinels.


lib/dialects/linux/dproc.c
```diff
@@ -94,7 +94,7 @@ struct l_fdinfo {
 
 #define EPOLL_MAX_TFDS 32
     int tfds[EPOLL_MAX_TFDS];
-    size_t tfd_count;
+    int tfd_count;
```
jiegec: I think `ssize_t` is better? Then the struct size is unchanged, so the ABI is compatible.

daeho-ro: I will try once I have a time 🫡

